PR #2, opened 2024-05-03
Add documentation for one tank example

ChristianWieck:
This MR:

- Adds a documentation page for the one tank example

ToDo's:

- [x] Complete the install instructions (espacially with hatch)
- [x] Add a reference to the 2024 ETFA paper once released (can be done in seperate MR)


docs/examples/one_tank.md
```diff
@@ -0,0 +1,219 @@
...
+It is inspired by the one-tank simulink model used for control design [^1].
+A schematic drawing of the system is shown in the following figure.
+
+![The one-tank system](./images/one_tank.svg)
+
+The complete source code for this example can be found [in the repository](https://github.com/flowcean/flowcean/blob/main/examples/one_tank/run.py).
+See also [Run this example](#run-this-example) on how to run this example locally.
+
+The dynamics of the system can be described by an ordinary differential equation (ODE).
+This type of equation relates the change in a variable (it's derivative) to its actual value.
```
newLabAspect, line 13: I know what you mean, but this sentence may be confusing. Propostion: "ODEs describe the derivative of a variable as a function of the variable itself."

docs/examples/one_tank.md
```diff
@@ -0,0 +1,219 @@
...
+
+For this example the system can be descirbed by the ODE
+
+$$ \dot{x} = \frac{b V(t) - a \sqrt{x}}{A} $$
+
+where $x$ is the water level in the tank, $\dot{x}$ is the change of the water level over time, $V(t)$ is the time-dependent inflow, $A$ is the tank area, and $a$ and $b$ are scaling constants for the equation.
+The solution of an ODE is not a single value, but a function (here $x(t)$) or a series of it's values for different times $t$.
```
newLabAspect, line 20: Should we make this less general and only say how it is in our case. I guess, that would be more clear.

docs/examples/one_tank.md
```diff
@@ -0,0 +1,219 @@
...
+As solving an ODE can be quite complicated, ofthen numerical solvers are used which compute solution points starting from an initial value.
+In this example, the initial value is the initial level of the liquid $x(0) = x_0$ in the tank.
+
+The other free parameters from the above equation are set to
```
newLabAspect, line 24: remove "other"

docs/examples/one_tank.md
newLabAspect: Also for line 21

docs/examples/one_tank.md
```diff
@@ -0,0 +1,219 @@
...
+
+ $A$ | $b$ |  $a$
+-----|-----|------
+ $5$ | $2$ | $0.5$
+
+The inflow is given by $V(t) = \mathrm{max}\left(0, \sin\left( 2 \pi \frac{1}{10} t \right)\right)$ and as the initial condition is $x_0 = 1$.
```
newLabAspect, line 30: remove "as"

docs/examples/one_tank.md
```diff
@@ -0,0 +1,219 @@
...
+Using a suitable numerical solution algorithm, the equation can be solved for the level $x_n$.
+Since the solution is not continuous, the level is not a function of time, but a discrete function of the sample number $n$.
+The corresponding time can be calculated by multiplying the sample number $n$ by the step size $h$ between two samples.
```
newLabAspect, line 33: to me, "T" is more common than "h" for the sampling period. Specifically as "h" may be confused with the height of the tank, here.

docs/examples/one_tank.md
```diff
@@ -0,0 +1,219 @@
...
+The graph below shows the development of the water level $x$ from zero to ten seconds.
+
+![Differential equation solution plotted over time](./images/one_tank_graph.svg)
+
+## Learning Modeles
```
newLabAspect, line 38: Type "Models"

docs/examples/one_tank.md
```diff
@@ -0,0 +1,219 @@
...
+
+After setting up the simulation, we want to use two different learners to predict the level of the tank $x[n]$ given the current input $V[n]$ and the level and input in the previous two time steps.
+The unknown function we are looking for and that we want to learn is
+
+$$ x_n = f\left(V_n, x_{n-1}, V_{n-1}, x_{n-2}, V_{n-2}\right). $$
+
+To do this, we first need data to learn the function from in Flowcean.
```
newLabAspect, line 45: "from in"?

docs/examples/one_tank.md
```diff
@@ -0,0 +1,219 @@
...
+        ],
+    ),
+    x0,
+    g=lambda t, x: np.array([x[0], V(t)]),
+    tstep=tstep,
+    output_names=["x", "V"],
+)
+```
+
+The output of the environment has the form
```
newLabAspect, line 72: What do you mean by "output"?

docs/examples/one_tank.md
```diff
@@ -0,0 +1,219 @@
...
+
+   $x$      |  $V$
+  ----------|-----------
+   $x[0]$   |  $V[0]$
+   $x[1]$   |  $V[1]$
+   $\dots$  |  $\dots$
+   $x[N]$   |  $V[N]$
+
+Since the learners we will use later only support learning on a fixed amount of data (called "offline learners" in the framework), we need to convert the incremental dataset into a fixed size dataset.
```
newLabAspect, line 81: refer to learning strategies in "concepts"?

docs/examples/one_tank.md
```diff
@@ -0,0 +1,219 @@
...
+This can be done by calling the [`take(N)`](../reference/flowcean/core/environment/incremental.md#flowcean.core.environment.incremental.IncrementalEnvironment.take) method on any `IncrementalEnvironment` to get $N$ samples and feed those into a [`Dataset`](../reference/flowcean/data/dataset.md).
+
+```python
+data = Dataset(data_incremental.load().take(250))
+```
+
+Until now, the data is in a time series format with each row representing a the values at a sample step $n$.
```
newLabAspect, line 88: "a the"?

docs/examples/one_tank.md
```diff
@@ -0,0 +1,219 @@
...
+```
+
+With the training data generated, fully transformed and split it's time to use learning algorithms to learn the prediction function from the beginning of this section.
+We use two different learners which both belong the the category of [offline learners](../user_guide/learning_strategies.md).
+
+First, a [regression tree](../reference/flowcean/learners/regression_tree.md) is used to learn a model.
+The implementation of this learner is part of the scikit-learn library.
+The learned model consists of a sequence of binary questions / comparisons that lead to the model result.
+The maximum depth, i.e. the number of questions asked on each path, is limited to five.
+The learners class is created and the helper method [`learn_offline`](../reference/flowcean/strategies/offline.md#flowcean.strategies.offline.learn_offline) called to start the training process
```
newLabAspect, line 112: Typo: "learners class"

docs/examples/one_tank.md
```diff
@@ -0,0 +1,219 @@
...
+    outputs,
+)
+```
+
+The `inputs` and `outputs` variables contain the names of the input and output fields in the `train' dataframe.
+
+Secondly a [multi-layer perceptron](../reference/flowcean/learners/lightning.md) is used to create a model.
+This type of model consists of a set of neurones arranged in layers which are connected with the previous layer.
+The value of each neuron is calculated by weighting and summing up the values of the neurons in the previous layer and applying a non-linear function; in this case a [leaky ReLU function](https://en.wikipedia.org/wiki/Rectifier_(neural_networks)#Leaky_ReLU).
+The result can be read from the last layers neuron values.
```
newLabAspect, line 129: "layers Neuron"?

docs/examples/one_tank.md
```diff
@@ -0,0 +1,219 @@
...
+This learners implementation uses the [lightning framework](https://lightning.ai/docs/pytorch/stable/) which is a high-level wrapper around the well known [PyTorch](https://pytorch.org/) library.
```
newLabAspect, line 130: "learners implementation"?

docs/examples/one_tank.md
```diff
@@ -0,0 +1,219 @@
...
+    train,
+    perceptron_learner,
+    inputs,
+    outputs,
+)
+```
+
+The final step is to evaluate the obtained models.
+This is done to estimate how well they are able to describe the unknown function as described above.
+Flowcean ships with a couple of different [metrices](../reference/flowcean/metrics/index.md) which can be used for this purpose.
```
newLabAspect, line 152: "metrices" -> "metrics" (I never saw the plural with an extra "e" before)

docs/examples/one_tank.md
```diff
@@ -0,0 +1,219 @@
...
+perceptron_report = evaluate(
+    perceptron_model,
+    test,
+    inputs,
+    outputs,
+    [MeanAbsoluteError(), MeanSquaredError()],
+)
+```
+
+For this example, the resulting metrics are about[^2]
```
newLabAspect, line 176: Why referencing the paper here? The numerical results are actually independent on whether they are published or not. Do we actually have to cite the paper with the example. As the paper refers to the current snapshot of the framework, I would rather put it (together with potential future references) on the starting page or some other separate place.

docs/examples/one_tank.md
```diff
@@ -0,0 +1,219 @@
...
+
+ Learner typ            | Runtime              | Mean Absolute Error | Mean Squared Error
+ -----------------------|----------------------|---------------------|--------------------
+ Regression Tree        | $15.5\: \mathrm{ms}$ | $0.0206$            | $0.0006$
+ Multi-layer Perceptron | $813\: \mathrm{ms}$  | $0.0639$            | $0.00054$
+
+Depending on the size of the dataset, the way the train and test set or split and shuffeld, the learners configuration and other random facts, these values may varry.
```
newLabAspect, line 183: "or" -> "are"

docs/examples/one_tank.md
```diff
@@ -0,0 +1,219 @@
...
+Afterwards you can navigate to the `examples` folder and run the examples.
+
+```sh
+cd examples/one_tank
+python run.py
+```
+
+!!! todo
+        Explain how to:
+            2. Run the example using hatch
```
newLabAspect, line 216: this is done now?

newLabAspect (commented):
Some general comments:
- the usage of "can" is often not needed / can be avoided
- you should run a spell checker on the document. I have the feeling that there are still a few typos.

Apart from that very nice and elaborate

docs/examples/one_tank.md
ChristianWieck: I agree with the comment for line 21 and fixed it accordingly. I'd like to keep the statement from line 20 to emphasize the point that an ODE has a function as its solution, not a single value. This is probably clear to anyone who has worked with ODEs before, but not everyone has.

ChristianWieck:
Thanks! I think I fixed most of the points.

newLabAspect (approved):
Looks fine to me now :)

Issue #15, opened 2024-05-06
Dataloader for Time Series Data with different sampling frequencies


mknitt:
Discussion 2024-04-01: Include in data loader for the moment. Might be transferred to a transform later.

mknitt:
This is supported at least for rosbags with #86

PR #23, opened 2024-05-06
Added transforms for up- and downsampling

mknitt:
Adds transforms and tests for up- and downsampling of time series data.

Required for #15

### DIFF
--- a/src/flowcean/transforms/downsample.py
+++ b/src/flowcean/transforms/downsample.py
@@ -1,0 +1,62 @@
+import logging
+from datetime import timedelta
+
+import polars as pl
+
+from flowcean.core import Transform
+
+logger = logging.getLogger(__name__)
+
+
+class Downsample(Transform):
+    """Downsample a time series.
+
+    Downsample a time series to a lower sampling rate using the polars method
+    "group_by_dynamic". The method can downsample to the first, last or mean
+    value of the group.
+
+    Args:
+        time_column (str): The name of the time column.
+
+        sampling_rate (str | timedelta): The sampling rate to downsample to.
+
+        value (str): The value to use for the downsampling. Can be "first",
+            "last" or "mean".
+
+
+    """
+
+    def __init__(
+        self,
+        time_column: str,
+        sampling_rate: str | timedelta,
+        value: str = "first",
+    ) -> None:
+        self.sampling_rate = sampling_rate
+        self.time_column = time_column
+        self.value = value
+
+    def transform(self, data: pl.DataFrame) -> pl.DataFrame:
+        logger.debug("Downsampling time series")
+
+        if self.value == "first":
+            data = data.group_by_dynamic(
+                self.time_column, every=self.sampling_rate
+            ).agg(
+                pl.col(x).first()
+                for x in data.columns
+                if x != self.time_column
+            )
+        elif self.value == "last":
+            data = data.group_by_dynamic(
+                self.time_column, every=self.sampling_rate
+            ).agg(
+                pl.col(x).last() for x in data.columns if x != self.time_column
+            )
+        elif self.value == "mean":
+            data = data.group_by_dynamic(
+                self.time_column, every=self.sampling_rate
+            ).agg(
+                pl.col(x).mean() for x in data.columns if x != self.time_column
+            )
+        return data
--- a/src/flowcean/transforms/upsample.py
+++ b/src/flowcean/transforms/upsample.py
@@ -1,0 +1,44 @@
+import logging
+from datetime import timedelta
+
+import polars as pl
+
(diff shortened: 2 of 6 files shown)


src/flowcean/transforms/downsample.py
```diff
@@ -0,0 +1,62 @@
...
+    "group_by_dynamic". The method can downsample to the first, last or mean
+    value of the group.
+
+    Args:
+        time_column (str): The name of the time column.
+
+        sampling_rate (str | timedelta): The sampling rate to downsample to.
+
+        value (str): The value to use for the downsampling. Can be "first",
+            "last" or "mean".
```
schmidma, line 24:
```suggestion
        time_column: The name of the time column.
        sampling_rate: The sampling rate to downsample to.
        value: The value to use for the downsampling. Can be "first", "last" or "mean".
```

src/flowcean/transforms/downsample.py
```diff
@@ -0,0 +1,62 @@
...
+
+class Downsample(Transform):
+    """Downsample a time series.
+
+    Downsample a time series to a lower sampling rate using the polars method
+    "group_by_dynamic". The method can downsample to the first, last or mean
+    value of the group.
+
+    Args:
+        time_column (str): The name of the time column.
```
schmidma, line 19: we usually call these "columns" _features_

src/flowcean/transforms/downsample.py
```diff
@@ -0,0 +1,62 @@
...
+        self.time_column = time_column
+        self.value = value
+
+    def transform(self, data: pl.DataFrame) -> pl.DataFrame:
+        logger.debug("Downsampling time series")
+
+        if self.value == "first":
+            data = data.group_by_dynamic(
+                self.time_column, every=self.sampling_rate
+            ).agg(
```
schmidma, line 45: You may be able to do this `group_by_dynamic` call once, as this is lazy in polars and then subsequently do the different aggregations afterwards.

Or maybe even better, first select the proper query for each version like
```
aggretation = pl.col(x)..... for x in ...
```
and then always do the
```
data = data.groupby().agg(aggregation)
```

src/flowcean/transforms/downsample.py
```diff
@@ -0,0 +1,62 @@
...
+                pl.col(x).last() for x in data.columns if x != self.time_column
+            )
+        elif self.value == "mean":
+            data = data.group_by_dynamic(
+                self.time_column, every=self.sampling_rate
+            ).agg(
+                pl.col(x).mean() for x in data.columns if x != self.time_column
```
schmidma, line 60: I think there is something in polars to select all columns except one. Maybe this https://docs.pola.rs/py-polars/html/reference/expressions/api/polars.exclude.html

src/flowcean/transforms/upsample.py
```diff
@@ -0,0 +1,44 @@
...
+    def transform(self, data: pl.DataFrame) -> pl.DataFrame:
+        logger.debug("Upsampling time series")
+        return (
+            data.upsample(
+                time_column=self.time_column,
+                every=self.sampling_rate,
+                offset=self.offset,
+            )
+            .interpolate()
+            .fill_null(strategy="forward")
```
schmidma, line 43: expose the strategy to the user by providing it as an argument?

src/flowcean/transforms/upsample.py
```diff
@@ -0,0 +1,44 @@
...
+    """Upsample a time series.
+
+    Upsample a time series to a higher sampling rate using the polars method
+    "upsample". The method can upsample to a higher sampling rate and fill
+    missing values using interpolation.
+
+    Args:
+        time_column (str): The name of the time column.
+        sampling_rate (str | timedelta): The sampling rate to upsample to.
+        offset (str | timedelta): The offset to use for the upsampling.
```
schmidma, line 21:
```suggestion
        time_column: The name of the time column.
        sampling_rate: The sampling rate to upsample to.
        offset: The offset to use for the upsampling.
```

src/flowcean/transforms/upsample.py
```diff
@@ -0,0 +1,44 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class Upsample(Transform):
+    """Upsample a time series.
+
+    Upsample a time series to a higher sampling rate using the polars method
+    "upsample". The method can upsample to a higher sampling rate and fill
+    missing values using interpolation.
```
schmidma, line 16:
```suggestion
    missing values using an interpolation strategy.
```

schmidma (changes requested):
some minor remarks and suggestions

mknitt:
Closing this PR because the functionality is already provided by #122

Issue #62, opened 2024-05-21
Handling of Time Series Data


newLabAspect:
Involves:
- [x] #42 
- [x] #41 
- [x] #40 
- [x] #39 
- [x] #38 
- [x] #34 

(Gitlab Issue 97, created by @newLabAspect)

newLabAspect:
@ChristianWieck _commented:_

Perhaps this could be solved by reintroducing metadata for each column? This would at least solve the problem for the machine learner.

newLabAspect:
@mknitt _commented:_

[@JanChristianWieck](https://github.com/JanChristianWieck) and I discussed this and we think a good solution would be to save the time series in multidimensional numpy arrays inside of polars dataframes.
```

| feature 1           | feature 2               |
| ------------------- | ----------------------- |
| np.array([[0,2,4],  | np.array([[0,1,2,3,4],  |
|           [3,5,4]]) |           [2,7,1,9,6]]) |      
|                     |                         |
| np.array([[0,2],    | np.array([[0,1,2],      |
|           [2,7]])   |           [1,3,2]])     |      
|                     |                         | 
```
Each time series would be structured like this:

`np.array([time, data])`

newLabAspect:
@ChristianWieck _commented:_

As discussed in the software meeting on 29.02. we move from `np.array` to `pl.List` as they are more performand. Only downside is, that only vectors are supported. The link between data and time vector needs to be implemented in a different way (e.g. hard-coded `<feature_name>` ->` t_<feature_name>` or metadata).

ChristianWieck:
As discussed at our meeting on 04.06.:
We want to use `pl.Structs` for storing time series, by storing the time vector in a field `t` and the associated value vector in the field `value`.

schmidma:
All individually mentioned issues for this meta-issue are resolved, closing this for now.

PR #81, opened 2024-06-13
Add Offline ROS Example

mknitt:
Adds
- RosbagEnvironment and tests
- run.py which uses the RosbagEnvironment and applies transforms
- example rosbag
- support for time arrays with np.datetime64 format in the MatchSamplingRate transform


tests/data/test_rosbag.py
```diff
@@ -0,0 +1,37 @@
...
+        data = environment.get_data()
+
+        assert data.columns == [
+            "j100_0000/amcl_pose.time",
+            "j100_0000/amcl_pose.pose.pose.position.x",
+            "j100_0000/amcl_pose.pose.pose.position.y",
+            "j100_0000/platform/odom.time",
+            "j100_0000/platform/odom.pose.pose.position.x",
+            "j100_0000/platform/odom.pose.pose.position.y",
+        ]
```
ChristianWieck, line 33: This is a pretty weak test, in my opinion - it only really tests whether the code runs, without making strong assertions about any results. We can keep it for now, but should extend it by something stronger soon. Maybe recording an artifical ros bag would be an idea which just contains a sine wave or something similar ?

src/flowcean/transforms/match_sampling_rate.py
ChristianWieck, line 1: This transform still uses the old way to express time series data. With regard to #62 we should use the "new" format here. That change however would be outside of the scope of this PR and should be done in a new PR.

src/flowcean/environments/rosbag.py
```diff
@@ -0,0 +1,84 @@
...
+        topics: dict[str, list[str]],
+        custom_msgs_path: str | Path | None = None,
+    ) -> None:
+        """Initialize the RosbagEnvironment.
+
+        Args:
+            path: Path to the rosbag.
+            topics: Dictionary of topics to load. The keys are the topic names
+                and the values are lists of keys inside each message that
+                shall be extracted.
```
ChristianWieck, line 36: I know what you mean, but that feels really difficult to grasp and use for a normal enduser. Can we find another, clearer way to express the same? Also it should probably be `field` instead of `keys`.

src/flowcean/environments/rosbag.py
```diff
@@ -0,0 +1,84 @@
...
+
+import polars as pl
+from rosbags.dataframe import get_dataframe
+from rosbags.highlevel import AnyReader
+from rosbags.typesys import Stores, get_types_from_msg, get_typestore
+
+from flowcean.core.environment import OfflineEnvironment
+
+
+def guess_msgtype(path: Path) -> str:
```
ChristianWieck, line 13: Make a static function of the `RosbagEnvironment`.

src/flowcean/environments/rosbag.py
```diff
@@ -0,0 +1,84 @@
...
+    """Guess message type name from path."""
+    name = path.relative_to(path.parents[2]).with_suffix("")
+    if "msg" not in name.parts:
+        name = name.parent / "msg" / name.name
+    return str(name)
```
ChristianWieck, line 18: It's not clear to me what this function does. Can you either add inline comments to explain your thinking or add an example of the functions input/output in the docstring?

src/flowcean/environments/rosbag.py
```diff
@@ -0,0 +1,84 @@
...
+                shall be extracted.
+            custom_msgs_path: Path to folder that contains custom ros message
+                descriptions.
+        """
+        self.path = Path(path)
+        self.topics = topics
+        self.data = pl.DataFrame()
+        self.typestore = get_typestore(Stores.ROS2_HUMBLE)
+
+        if custom_msgs_path:
```
ChristianWieck, line 45:
```suggestion
        if custom_msgs_path is not None:
```
Being explicit about what you check is better than implicitly checking for None.

src/flowcean/environments/rosbag.py
```diff
@@ -0,0 +1,84 @@
...
+            path: Path to the rosbag.
+            topics: Dictionary of topics to load. The keys are the topic names
+                and the values are lists of keys inside each message that
+                shall be extracted.
+            custom_msgs_path: Path to folder that contains custom ros message
+                descriptions.
+        """
+        self.path = Path(path)
+        self.topics = topics
+        self.data = pl.DataFrame()
```
ChristianWieck, line 42: Why are you initializing an empty dataframe here?

src/flowcean/environments/rosbag.py
```diff
@@ -0,0 +1,84 @@
...
+                )
+                df = pl.DataFrame(
+                    {col: [df[col].to_list()] for col in df.columns}
+                )
+                joined_df = pl.concat(
+                    [joined_df, df],
+                    how="horizontal",
+                )
+            self.data = joined_df
+        return self
```
ChristianWieck, line 80: The `load` method still uses the "old" timeseries format. I would recommend migrating it to the new format as discussed in #62.

pyproject.toml
```diff
@@ -18,6 +18,8 @@ dependencies = [
   "polars",
   "ruamel.yaml",
   "typing-extensions",
+  "rosbags",
+  "rosbags-dataframe",
```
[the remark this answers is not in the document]
ChristianWieck: For testing `flowcean[all]` is installed. `flowcean[ros]` was added to it, so it *should* work... You added the dependency to `flowcean[ros]` and the direct dependencies in the same commit (a144862ed5af9ef6d27134106521b2ddba588516). If you remove only the direct dependencies it should still pass.

src/flowcean/environments/ros_msgs/ParticleCloud.msg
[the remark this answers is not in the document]
mknitt: fixed

examples/ros_offline/test_rosbag/test_rosbag.db3.dvc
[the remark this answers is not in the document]
mknitt: done

src/flowcean/environments/rosbag.py
mknitt: done

src/flowcean/environments/rosbag.py
mknitt: Added example usage with input/output to docstring

src/flowcean/environments/rosbag.py
mknitt: done

src/flowcean/environments/rosbag.py
mknitt: Not sure. I replaced it with `self.data = None`

.dvc/config
```diff
@@ -1,6 +1,11 @@
...
-    remote = tuhh-cloud
+    remote = test-s3
 ['remote "tuhh-cloud"']
     url = webdavs://cloud.tuhh.de/public.php/webdav
     user = Q9WX2KGsC2oSQMD
+['remote "test-s3"']
+    url = s3://flowcean-dvc
+    endpointurl = http://192.168.1.226:9000
+    access_key_id = K6w5UbYW4Z0hANNUtGlC
+    secret_access_key = zUWBjuv8U9CXvKYdcFPBrKmwhTGnGfYeUWqyi2uN
```
ChristianWieck, line 11: That's probably the test s3 instance from @schmidma right? Do we want to keep it in here? Also you commited the access key...

optional/ros/nav2_msgs/msg/BehaviorTreeLog.msg
ChristianWieck, line 1: Just a general comment: Do we want to keep the message files in the main repository or move them to a seperate one? How are they delivered to a user who does not clone flowcean from sources but potentially installs it from PyPI?

.dvc/config
schmidma: Yes, this was just the test instance. These secret keys are not valid anymore. The server does not exist anymore ;)

examples/ros_offline/run.py
```diff
@@ -0,0 +1,48 @@
...
+        },
+    ) | Select(
+        features=[
+            "j100_0000/amcl_pose.pose.pose.position.x",
+            "j100_0000/platform/odom.pose.pose.position.x",
+        ]
+    )
+    transformed_data = transforms.transform(data)
+    print(data.head())
+    print(transformed_data.head())
```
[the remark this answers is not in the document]
mknitt: This will be implemented in the future once we support loading high-dimensional data such as laser scan data, image data, or particle cloud data.

optional/ros/nav2_msgs/msg/BehaviorTreeLog.msg
mknitt: I removed all msg files from our repository and will provide documentation on how to include the required msgs into the typestore

mknitt:
Due to lots of changes that came with the new time series format and transforms, I will close the PR and create a new branch that includes the newest changes in order to avoid merging. Nonetheless, I will consider the feedback that I got in this PR.

PR #85, opened 2024-06-25
Add StackEnvironment

ChristianWieck:
This PR:
- Add a new `StackEnvironment` to combine two (or more) `OfflineEnviroments` vertically
- Add helper methods and syntactic sugar to the `OfflineEnvironment`
- Add tests for the new environment

### DIFF
--- a/src/flowcean/core/environment/offline.py
+++ b/src/flowcean/core/environment/offline.py
@@ -49,14 +49,3 @@ class OfflineEnvironment(Environment):
         from .stack import StackEnvironment
 
         return StackEnvironment(self, other)
-
-    def __and__(self, other: OfflineEnvironment) -> OfflineEnvironment:
-        """Combine this environment with another one vertically.
-
-        Args:
-            other: The environment to append vertically.
-
-        Returns:
-            The combined environment.
-        """
-        return self.stack(other)
--- a/src/flowcean/core/environment/offline.py
+++ b/src/flowcean/core/environment/offline.py
@@ -35,3 +35,28 @@ class OfflineEnvironment(Environment):
             A streaming offline data.
         """
         return StreamingOfflineData(self, batch_size)
+
+    def stack(self, other: OfflineEnvironment) -> OfflineEnvironment:
+        """Combine this environment with another one vertically.
+
+        Args:
+            other: The environment to append vertically.
+
+        Returns:
+            The combined environment.
+        """
+        # Necessary to prevent circular imports
+        from .stack import StackEnvironment
+
+        return StackEnvironment(self, other)
+
+    def __and__(self, other: OfflineEnvironment) -> OfflineEnvironment:
+        """Combine this environment with another one vertically.
+
+        Args:
+            other: The environment to append vertically.
+
+        Returns:
+            The combined environment.
+        """
+        return self.stack(other)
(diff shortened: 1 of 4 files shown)


src/flowcean/core/environment/stack.py
```diff
@@ -0,0 +1,36 @@
...
+        """Get data from the environment.
+
+        Returns:
+            The loaded dataset.
+        """
+        dfs = [env.get_data() for env in self.environments]
+        return reduce(
+            lambda x, y: x.vstack(y),
+            dfs,
+        )
```
schmidma, line 36: There is [`pl.concat`](https://docs.pola.rs/api/python/stable/reference/api/polars.concat.html)

src/flowcean/core/environment/offline.py
```diff
@@ -35,3 +35,28 @@ def as_stream(self, batch_size: int = 1) -> StreamingOfflineData:
...
+    def __and__(self, other: OfflineEnvironment) -> OfflineEnvironment:
+        """Combine this environment with another one vertically.
+
+        Args:
+            other: The environment to append vertically.
+
+        Returns:
+            The combined environment.
+        """
+        return self.stack(other)
```
schmidma, line 62: I'm not sure about that. This is something like `a_env + b_env`, correct? This syntax is not showing whether it concatenates them vertically or horizontally. This syntax could also combine features.

src/flowcean/core/environment/stack.py
```diff
@@ -0,0 +1,36 @@
...
+
+from .offline import OfflineEnvironment
+
+
+class StackEnvironment(OfflineEnvironment):
+    """Combines multiple OfflineEnvironments into a single one.
+
+    This environment combines multiple OfflineEnvironments into a single one by
+    stacking them vertically. All environments need to share the identical
+    features.
```
schmidma, line 14: The fact that they must share identical feature names is only checked in the end, when the data is actually requested. Do we want to do that earlier? Or is this okay for now?

schmidma (changes requested):
minor questions

src/flowcean/core/environment/stack.py
ChristianWieck: Forgot about that, thanks

src/flowcean/core/environment/offline.py
ChristianWieck: The syntax is `env_a & env_b`. The general idea was to use `|` for horizontal concatenation of two frames. However, I understand that this is not clear. I would suggest to remove the feature for now and have a discussion about it in one of the next meetings?

src/flowcean/core/environment/stack.py
ChristianWieck: This boils down to a general question of error handling: Should we wait and "let the error happen"/polars tell us we did something wrong, or should we check when an environment/transformation is created and throw our own errors? Currently we mostly use the first approach. I'm open to both approaches as long as we keep one through the framework.

src/flowcean/core/environment/offline.py
schmidma: Agreed, sounds good

src/flowcean/core/environment/stack.py
schmidma: Let's keep it like this for now...

src/flowcean/core/environment/offline.py
ChristianWieck: I created the issue #89 to keep track of further discussions.

schmidma (approved):

PR #86, opened 2024-07-02
Add RosLoader to load offline data from rosbags

schmidma:
Adds a minimal example with ROS data from a single rosbag. This is a first step to supporting time series data. Transforms and further convenience methods will be added in coming PRs.


mknitt:
Looks good to me.

mknitt (approved):
Looks good to me.

PR #97, opened 2024-07-10
Documentation: Setting up Python

stbalduin:
I would not consider it finished, but more as a first draft. 

Please check it out, try it out and note down any issues you face.

### DIFF
--- a/docs/getting_started/installation.md
+++ b/docs/getting_started/installation.md
@@ -1,13 +1,42 @@
 # Installation
 
-It is recommended to use `flowcean` in a virtual environment.
-Consult [Preparation](preparation.md) or the official [python documentation](https://docs.python.org/3/library/venv.html) for further information.
+First, you have to download the source code of Flowcean (a package at PyPi will follow in the future).
+You can either use Git as recommended in the [Preparation](preparation.md) or download it as zip file from [Github](https://github.com/flowcean/flowcean).
 
-Install the flowcean package:
+If you followed the previous guide, you should have a folder `MyProject` which contains the virtual environment `myvenv`.
+Put the downloaded sources into the folder `MyProject` or exectute from within that folder:
+
+```bash
+git clone https://github.com/flowcean/flowcean.git
+cd flowcean
+```
+
+Your folder structure should now look like this:
+
+```
+|--MyProject/
+|  |--flowcean/
+|  |  |--docs/
+|  |  |--src/
+|  |  |--tests/
+|  |  |--pyproject.toml
+|  |  |--<other source files>
+|  |--myvenv/
+```
+
+It is recommended to use `flowcean` in a virtual environment.
+Make sure your virtual environment is [activated](preparation.md#virtual-environment) or consult the official [python documentation](https://docs.python.org/3/library/venv.html) for further information.
+
+Now you can install the Flowcean package (assuming that your current working directory is inside the `flowcean` folder):
 
 ```sh
 pip install -e .
 ```
 
-For the full functioning experience, flowcean splits its features into optional dependency packages.
+For the full functioning experience, Flowcean splits its features into optional dependency packages.
 Have a look at `pyproject.toml` and respectively install additional dependency groups.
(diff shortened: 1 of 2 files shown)


docs/getting_started/preparation.md
```diff
@@ -1,23 +1,147 @@
...
+
+```bash
+sudo apt-get install git  # On Debian/Ubuntu
+sudo pacman -S git  # On Arch linux
+```
 
 ## Python
 
-!!! todo
+Flowcean is written in Python and requires at least version 3.12.
```
schmidma, line 41: I think we only need hatch, and hatch will take care of installing the appropriate python version... And for hatch there are even standalone installers. Have a look at https://hatch.pypa.io/latest/install/

docs/getting_started/preparation.md
```diff
@@ -1,23 +1,147 @@
...
+Most linux distributions ship Python by default. 
+However, the version might be different from what is needed for Flowcean (>= 3.12).
+If you're on Debian/Ubuntu, you have to execute a sequence of commands:
+
+```bash
+sudo apt update && sudo apt upgrade -y
+sudo apt install software-properties-common -y
+sudo add-apt-repository ppa:deadsnakes/ppa
+sudo apt update && sudo apt install python3.12
+```
```
schmidma, line 100: Also here: we might just need any python or even better just `hatch`

docs/getting_started/preparation.md
```diff
@@ -1,23 +1,147 @@
...
+
+**TODO** Check if this works for "newer" versions
+
+Please check your distributions documentation for installation instructions if you're not on Debian/Ubuntu.
+
+Finally, you should be able to successfully check the installation as [described above](#python). 
 
 ## Virtual Environment
```
schmidma, line 108: Also virtual environments are fully managed by hatch. We do not need to do anything manually

docs/getting_started/installation.md
```diff
@@ -1,13 +1,42 @@
 # Installation
```
schmidma, line 1: I don't now whether we actually need any proper "Installation" if you work with flowcean, at the moment, using the hatch shell inside the cloned repository might be the best variant. As soon as we publish a proper release, we should also be pip-installable

schmidma (changes requested):

docs/getting_started/preparation.md
stbalduin: I have added a hint that hatch can be used and links to the documentation.  
The experienced user may use this and skip the rest of the preparation/installation.

For beginners, it is [advisable](https://www.bitecode.dev/p/why-not-tell-people-to-simply-use) to go the manual route, and we should provide the required steps here.

I don't use hatch and it works fine as well :-)

docs/getting_started/preparation.md
stbalduin: But in the pyproject file the version (>=3.12) is pinned, i.e., installation in a virtual environment will fail. (Regarding hatch, see my comment above)

docs/getting_started/preparation.md
schmidma: How do we ensure to always keep the python version in the documentation up-to-date? We already specify that in the `pyproject.toml` ... If we upgrade to a newer version, we have to ensure our python version in the documentation is always in sync with the documentation. This is why I preferred to mention hatch, as this respects the metadata in the `pyproject.toml`. If we keep the explanation for the manual way, I would be in favor of just linking to the python version, such that we are not in need of updating it.

docs/getting_started/preparation.md
schmidma: If you want to manually manage your python installations, this is a tedious task of finding out which python version works etc. Same as above, mentioning the versions in the documentation forces us to keep them in sync with the metadata in the code...

schmidma:
This is superseded by #167 in many parts. We should check, whether we can integrate some of this documentation ideas into the section on manual python installation.

newLabAspect:
Done by #275

PR #98, opened 2024-07-10
Match sampling rate reloaded

### LINKED ISSUE #40: TimeSeries - Match sampling transform
(Gitlab Issue 106, created by @ChristianWieck)

### LINKED ISSUE #62: Handling of Time Series Data

mknitt:
Closes #40, Closes #62

### DIFF
--- a/src/flowcean/transforms/match_sampling_rate.py
+++ b/src/flowcean/transforms/match_sampling_rate.py
@@ -1,5 +1,4 @@
 import logging
-from gc import collect
 from typing import Literal
 
 import polars as pl
--- a/src/flowcean/transforms/match_sampling_rate.py
+++ b/src/flowcean/transforms/match_sampling_rate.py
@@ -1,12 +1,14 @@
 import logging
+from typing import Literal
 
-import numpy as np
 import polars as pl
-from numpy import interp
+import polars.selectors as cs
 
 from flowcean.core.transform import Transform
 
 logger = logging.getLogger(__name__)
+
+type MatchSamplingRateMethod = Literal["linear"]
 
 
 class MatchSamplingRate(Transform):
@@ -17,14 +19,14 @@ class MatchSamplingRate(Transform):
     transform in a `run.py` file. Assuming the loaded data is
     represented by the table:
 
-    | feature_a                          | feature_b                          | const |
-    | ---                                | ---                                | ---   |
-    | list[struct[datetime[us],struct[]] | list[struct[datetime[us],struct[]] | int   |
-    | -----------------------------------| -----------------------------------| ----- |
-    | [{2024-06-25 12:26:01.0,{1.2}]},   | [{2024-06-25 12:26:00.0,{1.0}}, | 1     |
-    |  {2024-06-25 12:26:02.0,{2.4}]},   |  {2024-06-25 12:26:05.0,{2}}] |       |
-    |  {2024-06-25 12:26:03.0,{3.6}]},   |                                      |       |
-    |  {2024-06-25 12:26:04.0,{4.8}]}]   |                                      |       |
+    | feature_a                   | feature_b                   | const |
+    | ---                         | ---                         | ---   |
+    | list[struct[time,struct[]]] | list[struct[time,struct[]]] | int   |
+    | --------------------------- | --------------------------- | ----- |
+    | [{12:26:01.0, {1.2}},       | [{12:26:00.0, {1.0}},       | 1     |
+    |  {12:26:02.0, {2.4}},       |  {12:26:05.0, {2.0}}]       |       |
+    |  {12:26:03.0, {3.6}},       |                             |       |
+    |  {12:26:04.0, {4.8}}]       |                             |       |
 
     The following transform can be used to match the sampling rate
     of the time series `feature_b` to the sampling rate
@@ -35,9 +37,9 @@ class MatchSamplingRate(Transform):
         environment.load()
         data = environment.get_data()
         transform = MatchSamplingRate(
-            reference_timestamps="time_feature_a",
-            feature_columns_with_timestamps={
-                "feature_b": "time_feature_b"
+            reference_feature_name="feature_a",
+            feature_interpolation_map={
+                "feature_b": "linear",
             },
         )
         transformed_data = transform.transform(data)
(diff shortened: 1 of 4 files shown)


src/flowcean/transforms/match_sampling_rate.py
```diff
@@ -13,99 +13,166 @@ class MatchSamplingRate(Transform):
...
-    | [0, 1, 2]      | [2, 1, 7]   | [0, 1, 2]       | [10, 15, 20]    | 1    |
-    | [0, 1, 2]      | [4, 1, 0]   | [0, 1, 2]       | [20, 30, 40]    | 2    |
+    | feature_a                          | feature_b                          | const |
+    | ---                                | ---                                | ---   |
+    | list[struct[datetime[us],struct[]] | list[struct[datetime[us],struct[]] | int   |
+    | -----------------------------------| -----------------------------------| ----- |
+    | [{2024-06-25 12:26:01.0,{1.2}},    | [{2024-06-25 12:26:00.0,{1.0}},    | 1     |
+    |  {2024-06-25 12:26:02.0,{2.4}},    |  {2024-06-25 12:26:01.0,{1.2}},    |       |
+    |  {2024-06-25 12:26:03.0,{3.6}},    |  {2024-06-25 12:26:02.0,{2.4}},    |       |
+    |  {2024-06-25 12:26:04.0,{4.8}}]    |  {2024-06-25 12:26:03.0,{3.6}}]    |       |
```
schmidma, line 55: huh?! why is the time series not the same after interpolation? I'd assume that `feature_a` and `feature_b` have the same `datetime`s after interpolation?!

src/flowcean/transforms/match_sampling_rate.py
```diff
@@ -13,99 +13,166 @@ class MatchSamplingRate(Transform):
...
 
 
-
-    Note that the used feature `time_feature_b` is still
-    present in the DataFrame. To remove it use the `select` transform.
-    """
+    """  # noqa: E501
```
schmidma, line 58: Instead of silencing the warning here, I'd propose to shorten the the entries in the table. For example only showing the time and not the dates. They are the same anyway...

src/flowcean/transforms/match_sampling_rate.py
```diff
@@ -13,99 +13,166 @@ class MatchSamplingRate(Transform):
...
 
     def __init__(
         self,
-        reference_timestamps: str,
-        feature_columns_with_timestamps: dict[str, str],
+        reference_feature: str,
+        feature_columns: dict[str, str],
     ) -> None:
         """Initialize the MatchSamplingRate transform.
```
schmidma, line 65:
```suggestion
        """Initialize the transform.
```

This way, if we rename the class, we do not have to check for every occurrence in the docstrings.

src/flowcean/transforms/match_sampling_rate.py
```diff
@@ -13,99 +13,166 @@ class MatchSamplingRate(Transform):
...
 
         Args:
-            reference_timestamps: Timestamps of the reference feature.
-            feature_columns_with_timestamps: Names of the features that are
-                getting interpolated with their respective original timestamp
-                feature names.
+            reference_feature: Reference timeseries feature.
```
schmidma, line 68: Mention that this is the time series we target in interpolation.

src/flowcean/transforms/match_sampling_rate.py
```diff
@@ -13,99 +13,166 @@ class MatchSamplingRate(Transform):
...
+            feature_columns: key-value pairs of feature columns and
+                the interpolation method to use. The interpolation
+                method can be one of the following:
+                - 'linear'
+                - 'slerp' (to be implemented)
```
schmidma, line 73: If it is still to be implemented, just don't mention it

src/flowcean/transforms/match_sampling_rate.py
```diff
@@ -13,99 +13,166 @@ class MatchSamplingRate(Transform):
...
-    present in the DataFrame. To remove it use the `select` transform.
-    """
+    """  # noqa: E501
 
     def __init__(
         self,
-        reference_timestamps: str,
-        feature_columns_with_timestamps: dict[str, str],
+        reference_feature: str,
+        feature_columns: dict[str, str],
```
schmidma, line 63: I do not like the names. They are both neither features nor columns, they are just the names of the features. But that is fine, if we consider a feature to be a name to some data behind it.

I'd like to have the fact that many features are interpolated to a single reference time series represented in the naming.

src/flowcean/transforms/match_sampling_rate.py
```diff
@@ -13,99 +13,166 @@ class MatchSamplingRate(Transform):
...
     def transform(self, data: pl.DataFrame) -> pl.DataFrame:
+        """Transform the input DataFrame.
+
+        Args:
+            data: Input DataFrame.
+
+        Returns:
+            Transformed DataFrame.
+
+        """
```
schmidma, line 87: This is an override of a method of a parent class. Mark it as such and drop the docstring.
```suggestion
    @override
    def transform(self, data: pl.DataFrame) -> pl.DataFrame:
```

src/flowcean/transforms/match_sampling_rate.py
```diff
@@ -13,99 +13,166 @@ class MatchSamplingRate(Transform):
...
+        # preserve all constant columns that are not timeseries data
+        non_timeseries_features = [
+            col
+            for col in data.columns
+            if col not in [*self.feature_columns, self.reference_feature]
+        ]
```
schmidma, line 93: You can use the polars expression syntax to express that you want to have everything except some columns. Have a look at the column selectors for example.

src/flowcean/transforms/match_sampling_rate.py
```diff
@@ -13,99 +13,166 @@ class MatchSamplingRate(Transform):
...
         logger.debug("Matching sampling rate of time series.")
```
schmidma, line 94: We should give some more relevant information in this log. Like the features we are matching etc.

src/flowcean/transforms/match_sampling_rate.py
```diff
@@ -13,99 +13,166 @@ class MatchSamplingRate(Transform):
...
-                        pl.when(pl.col("index") == i)
-                        .then(resampled_timeseries)
-                        .otherwise(pl.col(feature))
-                        .alias(feature)
-                    )
-                    .drop("index")
-                    .collect()
+        if self.reference_feature not in data.columns:
+            msg = f"Reference {self.reference_feature} not found in the DataFrame."  # noqa: E501
+            raise ValueError(msg)
```
schmidma, line 98: You can reformat the string or even better create a new exception class that properly describes your error. Something similar to the `NotLoadedError`

src/flowcean/transforms/match_sampling_rate.py
```diff
@@ -13,99 +13,166 @@ class MatchSamplingRate(Transform):
...
+
+        features = list(self.feature_columns.keys())
```
schmidma, line 100:
```suggestion
        features = self.feature_columns.keys()
```

src/flowcean/transforms/match_sampling_rate.py
```diff
@@ -13,99 +13,166 @@ class MatchSamplingRate(Transform):
...
+        reference_df = (
+            data.select(pl.col(self.reference_feature).explode())
```
schmidma, line 102: Wait a second... Does this work if we have multiple entries in our time series feature? Up until now you only considered dataframes with a single row...

src/flowcean/transforms/match_sampling_rate.py
```diff
@@ -13,99 +13,166 @@ class MatchSamplingRate(Transform):
...
+            .unnest(cs.all())
+            .rename(
+                lambda name: self.reference_feature + "_" + name
+                if name != "time"
+                else name
+            )
```
schmidma, line 108: We always consider these to be columns named `time` and `value` don't we?

src/flowcean/transforms/match_sampling_rate.py
```diff
@@ -13,99 +13,166 @@ class MatchSamplingRate(Transform):
...
+            [
+                data.select(self.reference_feature),
+                result,
+                data.select(non_timeseries_features),
+            ],
+            how="horizontal",
+        )
+
+    def interpolate_feature(
+        self, feature: str, data: pl.DataFrame, reference_df: pl.DataFrame
```
schmidma, line 128:
```suggestion
        self,
        feature: str,
        data: pl.DataFrame,
        reference_df: pl.DataFrame,
```

schmidma, line 128: Do not repeat yourself (DRY)
```suggestion
        self, feature: str, data: pl.DataFrame, reference: pl.DataFrame
```
Do not mention types in variable names

src/flowcean/transforms/match_sampling_rate.py
```diff
@@ -13,99 +13,166 @@ class MatchSamplingRate(Transform):
...
+        return pl.concat(
+            [
+                data.select(self.reference_feature),
+                result,
+                data.select(non_timeseries_features),
+            ],
+            how="horizontal",
+        )
+
+    def interpolate_feature(
```
schmidma, line 127: The only reason this function is a member function taking `self` is to access the interpolation method. I'd suggest to move this function to a private free function and also pass the interpolation method as an argument.

schmidma (changes requested):
Some first comments... I'll have a deeper look at the algorithm itself later

src/flowcean/transforms/match_sampling_rate.py
mknitt: they do have the same `datetime`s. I fixed the error in the docstring

src/flowcean/transforms/match_sampling_rate.py
mknitt: Thanks, I removed the dates, only keeping the times

src/flowcean/transforms/match_sampling_rate.py
mknitt: i removed it

src/flowcean/transforms/match_sampling_rate.py
mknitt: doen

src/flowcean/transforms/match_sampling_rate.py
mknitt: I changed the variable names to `reference_feature_name` and `feature_interpolation_map`.

src/flowcean/transforms/match_sampling_rate.py
mknitt: accepted the suggestion

src/flowcean/transforms/match_sampling_rate.py
mknitt: done

src/flowcean/transforms/match_sampling_rate.py
mknitt: Added some more information to the debug message

src/flowcean/transforms/match_sampling_rate.py
mknitt: done

src/flowcean/transforms/match_sampling_rate.py
mknitt: I added a test for 2-dimensional values and it works

src/flowcean/transforms/match_sampling_rate.py
mknitt: yes, this would be an example for the structure that we agreed on: 
```
"feature_b": [
      [
          {
              "time": datetime(
                  2024, 6, 25, 12, 26, 1, 0, tzinfo=UTC
              ),
              "value": {"feature_b_x": 1.2, "feature_b_y": 12.0},
          },
          {
              "time": datetime(
                  2024, 6, 25, 12, 26, 2, 0, tzinfo=UTC
              ),
              "value": {"feature_b_x": 1.4, "feature_b_y": 14.0},
          },
      ]
],
```
We did not agree on limitations regarding the naming of keys inside the content of the `value` key, did we?

src/flowcean/transforms/match_sampling_rate.py
mknitt: changed

mknitt: renamed the variable

src/flowcean/transforms/match_sampling_rate.py
mknitt: done

src/flowcean/transforms/match_sampling_rate.py
schmidma: I think I meant having a dataframe with multiple time series. Currently you only considered a single time series (thus a single row) in the original dataframe. What happens if you have multiple runs, multiple rows? Not multiple dimensions (that means columns) in a single time series, but multiple rows?

src/flowcean/transforms/match_sampling_rate.py
schmidma: I meant, you provide a lambda knowing, that you only ever will see a column named `value` and the `time` column you already excluded with the if. So is there actually a need for the lambda function?

src/flowcean/transforms/match_sampling_rate.py
```diff
@@ -13,99 +14,187 @@ class MatchSamplingRate(Transform):
...
+                *self.feature_interpolation_map.keys(),
+                self.reference_feature_name,
+            )
+        )
+        debug_msg = (
+            "Interpolating timeseries features: "
+            + str(self.feature_interpolation_map.keys())
+            + " using the timestamps of the reference feature "
+            + self.reference_feature_name
+        )
```
schmidma, line 91: You can use format strings for that. something like
```python
f"Interpolating timeseries features: {self.feature_interpolation_map.keys()} using the timestamps of the reference feature {self.reference_feature_name}"

```

src/flowcean/transforms/match_sampling_rate.py
mknitt: Before it did not, but now it does work now for timeseries dataframes with multiple rows.

src/flowcean/transforms/match_sampling_rate.py
mknitt: I changed the input to the rename function, using a dict instead of the lambda function

src/flowcean/transforms/match_sampling_rate.py
mknitt: switched to f-strings.

src/flowcean/transforms/match_sampling_rate.py
```diff
@@ -46,87 +45,166 @@ class MatchSamplingRate(Transform):
...
-    present in the DataFrame. To remove it use the `select` transform.
-    """  # noqa: E501
+    """
 
     def __init__(
         self,
-        reference_timestamps: str,
-        feature_columns_with_timestamps: dict[str, str],
+        reference_feature_name: str,
+        feature_interpolation_map: dict[str, str],
```
ChristianWieck: Replace `str` with an `InterpolationMethodType` to constraint the choice here. This should make type checking easier and provide better hints.

mknitt:
> Some first comments... I'll have a deeper look at the algorithm itself later

All comments have been addressed. Do you still want to take a deeper look at the algorithm?

schmidma:
The diff of this branch looks weird... can you rebase this PR to the latest main? Thanks

src/flowcean/transforms/match_sampling_rate.py
```diff
@@ -46,89 +48,167 @@ class MatchSamplingRate(Transform):
...
+                                if col != "time"
+                            },
+                        ).alias("value"),
+                    ).alias(target_feature_name),
+                ],
+            )
+            .select(pl.all().implode())
+        )
+    msg = f"Unknown interpolation method: {interpolation_method}"
+    raise ValueError(msg)
```
schmidma, line 207: Create an error for this as well?

src/flowcean/transforms/match_sampling_rate.py
```diff
@@ -46,89 +48,167 @@ class MatchSamplingRate(Transform):
...
+        debug_msg = (
+            "Interpolating timeseries features: "
+            f"{self.feature_interpolation_map.keys()} using the timestamps of "
+            f"the and reference feature {self.reference_feature_name}"
+        )
+        logger.debug(debug_msg)
+
+        if self.reference_feature_name not in data.columns:
+            msg = f"'{self.reference_feature_name}' not in the DataFrame."
+            raise FeatureNotFoundError(msg)
```
schmidma, line 114: I'd rather pass only the name of the feature, and construct the error message in the constructor of `FeatureNotFoundError`
something like:
```python
class FeatureNotFoundError(Exception):
    def __init__(self, feature: str) -> None:
        super().__init__(f"{feature} not found")
```

src/flowcean/transforms/match_sampling_rate.py
```diff
@@ -46,89 +48,167 @@ class MatchSamplingRate(Transform):
...
+
+        Returns:
+            Transformed DataFrame.
+
+        """
+        # preserve all constant columns that are not timeseries data
+        transformed_data = pl.DataFrame()
+        for i in range(len(data.collect().rows())):
+            transformed_data_slice = self._transform_row(
+                data.collect().slice(i, 1),
```
schmidma, line 93: you collect twice here...

src/flowcean/transforms/match_sampling_rate.py
```diff
@@ -46,89 +48,167 @@ class MatchSamplingRate(Transform):
...
+        Args:
+            data: Input DataFrame.
+
+        Returns:
+            Transformed DataFrame.
+
+        """
+        # preserve all constant columns that are not timeseries data
+        transformed_data = pl.DataFrame()
+        for i in range(len(data.collect().rows())):
```
schmidma, line 91: Iterating rows might be extremely slow... If it is fine for now, we can keep it like that. Alternatively we should try to use the expression syntax, if that is possible

schmidma, line 91: Consider using `.map_rows()`

src/flowcean/transforms/match_sampling_rate.py
```diff
@@ -46,89 +48,167 @@ class MatchSamplingRate(Transform):
...
+                for feature in features
+            ],
+            how="horizontal",
+        )
+
+        return pl.concat(
+            [
+                data.select(self.reference_feature_name),
+                result,
+                data.select(non_timeseries_features),
```
schmidma, line 141: This already is a selected expression, isn't it?

```suggestion
                non_timeseries_features,
```

src/flowcean/transforms/match_sampling_rate.py
```diff
@@ -46,89 +48,167 @@ class MatchSamplingRate(Transform):
...
+            f"{self.feature_interpolation_map.keys()} using the timestamps of "
+            f"the and reference feature {self.reference_feature_name}"
+        )
+        logger.debug(debug_msg)
+
+        if self.reference_feature_name not in data.columns:
+            msg = f"'{self.reference_feature_name}' not in the DataFrame."
+            raise FeatureNotFoundError(msg)
+
+        features = list(self.feature_interpolation_map.keys())
```
schmidma, line 116:
```suggestion
        features = self.feature_interpolation_map.keys()
```

This does not have to be a list, does it?

src/flowcean/transforms/match_sampling_rate.py
```diff
@@ -46,89 +48,167 @@ class MatchSamplingRate(Transform):
...
+                            {
+                                col: pl.col(col)
+                                for col in feature_df.columns
+                                if col != "time"
+                            },
+                        ).alias("value"),
+                    ).alias(target_feature_name),
+                ],
+            )
+            .select(pl.all().implode())
```
schmidma, line 204: I'd suggest to split this into multiple assignments and give them appropriate names. This way, the final chain is much more readable.

src/flowcean/transforms/match_sampling_rate.py
mknitt: done

src/flowcean/transforms/match_sampling_rate.py
mknitt: created another variable instead

src/flowcean/transforms/match_sampling_rate.py
mknitt: I tried, but could not make it work. Maybe we can postpone this.

src/flowcean/transforms/match_sampling_rate.py
mknitt: done

src/flowcean/transforms/match_sampling_rate.py
mknitt: changed

src/flowcean/transforms/match_sampling_rate.py
mknitt: done

schmidma (approved):

PR #100, opened 2024-07-10
Docu Update

newLabAspect:
- minor updates User Guide
- remove outdated description Coffee Machine
- replace outdated description Failure Time Prediction
- minor corrections One Tank

### DIFF
--- a/docs/examples/coffee_machine.md
+++ b/docs/examples/coffee_machine.md
@@ -1,20 +1,23 @@
 # Coffee Machine Example
 
-This example shows how to train a model to predict the behavior of an automaton. 
+This example shows how to train a model to predict the behavior of an automaton.
 Inspired by the work of Steffen et al.[^1], we consider a simple coffee machine.
 LearnLib[^2], a framework for automata learning written in Java, is used for the inference of the automaton.
 
 !!! todo
 
-    This explanation is outdated. The automaton example has to be refined using time series data.
+    This explanation is outdated.
+    The automaton example has to be refined using time series data.
 
 ## Run this example
 
 ### Docker Container
 
-The external learner written in Java is containerized for this example. Thus, Docker must be installed to run it.
+The external learner written in Java is containerized for this example.
+Thus, Docker must be installed to run it.
 Follow the instructions on the [official Docker page](https://docs.docker.com/get-docker/) to install it.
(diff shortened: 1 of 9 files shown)


docs/examples/coffee_machine.md
```diff
@@ -1,57 +1,31 @@
...
-learners:
-  - name: Automaton Learner
-    class_path: agenc.learners.grpc.GrpcLearner
-    arguments:
-      port: 8080
-      ip: "localhost"
-```
+## Local Execution
 
+It is also possible to directly run the Java server locally.
```
schmidma, line 25: Please put one line per sentence. This makes line based version diff much easier to read and maintain

docs/examples/failure_time_prediction.md
```diff
@@ -30,20 +30,30 @@ The `preprocessing.ipynb` notebook is used to reproduce the steps and generate a
...
+
+### From source
+
+Follow the [installation guide](../getting_started/installation.md) to install flowcean and it's dependencies from source.
+Afterwards you can navigate to the `examples` folder and run the examples.
+
+```sh
+cd examples/failure_time_prediction
+python run.py
+```
```
schmidma, line 59: Is this a proper alternative? I'd just use `hatch shell`. Then hatch again takes care of setting things up, including installing dependencies etc.

docs/user_guide/environment.md
```diff
@@ -28,7 +28,7 @@ classDiagram
   }
   class IncrementalEnvironment{
     Loads data in an iterative way. 
-    + get_next_data() -> DataFrame
+    + collect() -> DataFrame
```
schmidma, line 31:
```suggestion
    + collect(n: int) -> DataFrame
```

schmidma (changes requested):

docs/examples/failure_time_prediction.md
newLabAspect: As we explain how to install flowcean yourself, I like having the explanation how to run the example in your current python environment. We also mention that using `hatch run` is the easier option.

But I am also fine with removing this, if you consider it confusing or unnecessary.

docs/examples/coffee_machine.md
newLabAspect: Done. Also for all (if I didn't miss something) other docu files.

PR #102, opened 2024-07-23
Euclidean distance

mknitt:
Adds a transform that can calculate the Euclidean Distance for time series features. 

Usage is shown in the ros_offline example.

This is based on match sampling rate reloaded branch (#98)


src/flowcean/transforms/match_sampling_rate.py
ChristianWieck, line 1: Can we remove this file from the PR? I think there is a seperate PR ongoing where those changes are introduced.

tests/transforms/test_match_sampling_rate.py
ChristianWieck, line 1: Same as with `match_sampling_rate.py` - I think we should discuss the changes in it's own PR. However the `euclidean_distance` is missing test?

src/flowcean/transforms/euclidean_distance.py
```diff
@@ -0,0 +1,121 @@
...
+    Computes the Euclidean distance between two time series. The below example
+    shows the usage of an `EuclideanDistance` transform in a `run.py` file.
+    Assuming the loaded data is represented by the table:
+
+    | feature_a            | feature_b            |
+    | ---                  | ---                  |
+    | [{1.0, {1.0, 1.0}},  | [{1.0, {1.0, 2.0}},  |
+    |  {2.0, {2.0, 2.0}},  |  {2.0, {1.0, 2.0}},  |
+    |  {3.0, {3.0, 3.0}},  |  {3.0, {2.0, 3.0}},  |
+    |  {4.0, {4.0, 4.0}}]  |  {4.0, {3.0, 4.0}}]  |
```
ChristianWieck, line 24: What is the datatyype of the features value? Is a simply a list?

src/flowcean/transforms/euclidean_distance.py
```diff
@@ -0,0 +1,121 @@
...
+            .select(
+                [
+                    pl.struct(
+                        pl.col("time"),
+                        pl.struct(pl.col("value")),
+                    ).alias(self.output_feature_name)
+                ]
+            )
+            .select(cs.all().implode())
+        )
```
ChristianWieck, line 121: Just out of curiosity: Have you benchmarked the code? I think it's much cleaner than the current implementation of `resample`. If it is indeed faster, we might want to apply similar logic to the resample transformation.

src/flowcean/transforms/match_sampling_rate.py
mknitt: Good point, I removed it.

tests/transforms/test_match_sampling_rate.py
mknitt: I removed `match_sampling_rate.py` and added a test for `euclidean_distance`. It still needs some debugging, though.

PR #109, opened 2024-07-30
Add Time series flatten transform

### LINKED ISSUE #41: TimeSeries - Flatten transform
(Gitlab Issue 104, created by @ChristianWieck)

ChristianWieck:
This PR:
- Adds a new transform to flatten time series from the "new" format in a polars way.

Closes #41.

### DIFF
--- a/src/flowcean/transforms/flatten.py
+++ b/src/flowcean/transforms/flatten.py
@@ -1,0 +1,103 @@
+import logging
+from collections.abc import Iterable
+from typing import cast, override
+
+import polars as pl
+
+from flowcean.core import Transform
+
+logger = logging.getLogger(__name__)
+
+
+class Flatten(Transform):
+    """Flatten all time series in a DataFrame to individual features.
+
+    The given DataFrame's time series are converted into individual features,
+    with each time step creating a new feature. This transform will change the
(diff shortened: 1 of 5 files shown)


src/flowcean/transforms/flatten.py
```diff
@@ -0,0 +1,103 @@
...
+
+    gets flattend into the dataset
+
+    series_data_0 | series_data_1 | series_data_2 | A  | B
+    --------------|---------------|---------------|----|----
+    0             | 1             | 2             | 42 | 43
+    3             | 4             | 5             | 42 | 43
+    """
+
+    def __init__(self, features: Iterable[str] | None = None) -> None:
```
schmidma, line 34: This function is missing documentation and also the fact that `None` leads to automatic selection.

schmidma (changes requested):
one minor thing...

src/flowcean/transforms/flatten.py
ChristianWieck: Thanks for noting. Should be fixed now with an explicit mention that `None` selects all possible features.

schmidma (approved):

PR #110, opened 2024-07-30
Directory loader

ChristianWieck:
This PR:
- Adds a `load_directory` helper function to load `OfflineEnvironments` from a folder with a given `loader_function`
- Implements a `loader_function` for the csv environment

This PR is experimental/controversial. We should discuss it in the team before thinking about merging it.

ToDos:
- [x] Replace `Callable[[Path], OfflineEnvironment]` with a type alias for better readibility
- [x] Implement a loader function for the other oflfine environments
- [x] Add progress logging -> Moved into seperate PR.

### DIFF
--- a/src/flowcean/environments/csv.py
+++ b/src/flowcean/environments/csv.py
@@ -1,5 +1,4 @@
 import logging
-from collections.abc import Callable
 from pathlib import Path
 from typing import Self, override
 
@@ -7,6 +6,7 @@ import polars as pl
 
 from flowcean.core import OfflineEnvironment
 from flowcean.core.environment import NotLoadedError
+from flowcean.environments.directory_loader import LoaderFunction
 
 logger = logging.getLogger(__name__)
 
@@ -44,5 +44,5 @@ class CsvDataLoader(OfflineEnvironment):
         return self.data
 
     @staticmethod
-    def loader() -> Callable[[Path], OfflineEnvironment]:
+    def loader() -> LoaderFunction:
         return lambda p: CsvDataLoader(p).load()
--- a/src/flowcean/environments/csv.py
+++ b/src/flowcean/environments/csv.py
@@ -1,5 +1,4 @@
 import logging
-from collections.abc import Callable
 from pathlib import Path
 from typing import Self, override
 
@@ -42,7 +41,3 @@ class CsvDataLoader(OfflineEnvironment):
         if self.data is None:
             raise NotLoadedError
         return self.data
-
-    @staticmethod
-    def loader() -> Callable[[Path], OfflineEnvironment]:
-        return lambda p: CsvDataLoader(p).load()
--- a/src/flowcean/environments/csv.py
+++ b/src/flowcean/environments/csv.py
@@ -1,4 +1,5 @@
 import logging
+from collections.abc import Callable
 from pathlib import Path
 from typing import Self, override
 
@@ -41,3 +42,7 @@ class CsvDataLoader(OfflineEnvironment):
         if self.data is None:
             raise NotLoadedError
         return self.data
+
+    @staticmethod
+    def loader() -> Callable[[Path], OfflineEnvironment]:
+        return lambda p: CsvDataLoader(p).load()
--- a/src/flowcean/environments/csv.py
+++ b/src/flowcean/environments/csv.py
@@ -1,4 +1,5 @@
 import logging
+from collections.abc import Callable
 from pathlib import Path
 from typing import Self, override
 
@@ -6,7 +7,6 @@ import polars as pl
 
 from flowcean.core import OfflineEnvironment
 from flowcean.core.environment import NotLoadedError
-from flowcean.environments.directory_loader import LoaderFunction
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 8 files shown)


src/flowcean/environments/csv.py
```diff
@@ -41,3 +42,7 @@ def get_data(self) -> pl.DataFrame:
         if self.data is None:
             raise NotLoadedError
         return self.data
+
+    @staticmethod
+    def loader() -> LoaderFunction:
+        return lambda p: CsvDataLoader(p).load()
```
schmidma, line 48: Why do you need this wrapping function? The environment `__init__` methods already suffice the call signature of the `load_directory` function. Couldn't you just pass `CsvEnvironment` as an argument to `load_function`? Am I missing something?

src/flowcean/environments/directory_loader.py
```diff
@@ -0,0 +1,40 @@
...
+        include_folders: Specify whether to pass folders to the
+            `load_function`.
+    """
+    for item in Path(path).glob(pattern):
+        if (item.is_file() and include_files) or (
+            item.is_dir() and include_folders
+        ):
+            environment = load_function(item)
+            if environment is not None:
+                yield environment
```
schmidma, line 40: Not sure whether this actually is easier to read/understand, but I am a fan of functional style expressions in this case. Something like:

```python
map(
  load_function,
  filter(
    lambda item: item.is_file() or item.is_dir(),
    Path(path).glob(pattern)
  )
)
```

You can additionally extract the intermediate iterables into variables to increase readability and give it more structure.

src/flowcean/environments/csv.py
ChristianWieck: While the different offline envrionments share a similar `__init__` it is nowhere explicitly defined / inherited from the base class. I could ofcourse just *assume* that i can pass a path into an environment and then call `load` on it, but that's setup for failure if an environment has a different `__init__` method...
That's the reason for the `loader` function: To have a clear interface to create an environment from a path (if it's supported)...
With that I'm not sure if you can pass a (typed) constructor to another method in python...

src/flowcean/environments/directory_loader.py
ChristianWieck: I would personally keep it in the "long" form - espacially because we can easily add a progress ar here (which I did in #128). If we can add the progressbar to the function style expression as well, I'm all in for it 😄

src/flowcean/environments/csv.py
schmidma: You can.

The interface is already defined by the `load_directory` function. This expects a function that takes a single argument (a path) to load an environment. If the constructor "accidentally" fits this signature, why introduce another wrapping `load` function?! Even without changing the interface of the `load_directory` function, the constructor already suffices that interface.

src/flowcean/environments/directory_loader.py
schmidma: The tqdm just takes an `Iterable` and produces another. So, yes, you can.

ChristianWieck: Then the functional style is fine for me. I updated the code accordingly.

src/flowcean/environments/csv.py
ChristianWieck: I didn't know that you can just pass classes/the constructors around like this in python... It's tbh the first language I use that allows such things.
In that case the loader function is indeed superfluous an can be removed.
Shall we call `Load` from within the `load_directory` method? Otherwise the user has to write a loop just to call `load` on all environments.

schmidma: Is it necessary to load them at this stage? I think the helper function should not yet load the data.

ChristianWieck: By definition we need to `load` the data before using it (e.g. passing it through a `TestTrainSplit`). In all our example codes we create the environment and directly afterwards call `load`. I would personally argue that `load` is not needed at all as part of the public interface and should be called once by `get_data`, but that's out-of-scope for this PR.
Right now you would need to loop through all environments and call `load` on them manually...

src/flowcean/environments/directory_loader.py
```diff
@@ -0,0 +1,44 @@
+from collections.abc import Callable, Iterable
+from pathlib import Path
+
+from flowcean.core.environment.offline import OfflineEnvironment
+
+LoaderFunction = Callable[[Path], OfflineEnvironment]
+
+
+def load_directory(
```
schmidma, line 9: I'm not sure whether this function does what it says. At least for now it does not load anything (and that's a good thing). In the end it just maps a function to a list of paths... I feel like we might not even need that function in the end.

The only functionality it provides is filtering for files or directories. But this is also something I doubt is designed wisely: What about symbolic links for example?

Without such a function the user would just write:
```python
environments = map(CsvEnvironment, Path("my_data").glob("*.csv"))
```
This is something we are able to do in `main` and at least to my eyes this is already simple enough.

TLDR, I'd suggest to not provide such a function because it is not actually doing anything...
What do you think?

src/flowcean/environments/csv.py
schmidma: But usually the user would call some variant of a stack or combine transform anyway?! And then it is a single environment with a single load function... I currently really like that we split initialization from loading data. This extremely speeds up early preprocessing of data and allows to check for structural bugs early on before even loading any data.

schmidma (approved):

schmidma (approved):

PR #113, opened 2024-07-31
Add `to_time_series` method

ChristianWieck:
This PR:
- Adds a `to_time_series` method to `OfflineEnvironment` to convert one to a single entry in a time series envrionment.

For now this method is directly attached to the `OfflineEnvironment`. This however adds "production" code to the base class. We can / should discuss if and where we could move such a method instead.

### DIFF
--- a/src/flowcean/core/environment/offline.py
+++ b/src/flowcean/core/environment/offline.py
@@ -67,7 +67,6 @@ class OfflineEnvironment(Environment):
 
         # Get the underlaying dataframe
         data = self.get_data()
-        result_data = pl.DataFrame()
         # Create the time feature mapping
         if isinstance(time_feature, str):
             time_feature = {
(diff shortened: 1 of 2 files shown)


src/flowcean/core/environment/offline.py
```diff
@@ -49,3 +47,53 @@ def stack(self, other: OfflineEnvironment) -> OfflineEnvironment:
...
+                .rename(
+                    {
+                        t_feature: "time",
+                        value_feature: "value",
+                    }
+                )
+                .to_struct()
+                .implode()
+                .alias(value_feature)
+            )
```
schmidma, line 97: You may be able to replace this with a single polars expression...
For example something like this:

```python
data = pl.DataFrame(
    {
        "time_feature": [0, 1, 2, 3],
        "feature_a": [42, 43, 44, 45],
        "feature_b": [3, 2, 1, 0],
    }
)

res = data.select(
    [
        pl.struct(
            pl.col("time_feature").alias("time"),
            pl.col(value_feature).alias("value"),
        )
        .implode()
        .alias(value_feature)
        for value_feature in ["feature_a", "feature_b"]
    ]
)
```

schmidma (changes requested):

src/flowcean/core/environment/offline.py
ChristianWieck: That's great. I didn't know about the `pl.struct` method to create structs directly.

schmidma (approved):

src/flowcean/core/environment/offline.py
ChristianWieck: Can you resolve the conversation so we can merge this one? 😅

PR #115, opened 2024-08-01
Add Joined environment

ChristianWieck:
This PR:
- Adds a `CombineEnvironment` which combines the features of multiple environments into a single one.
- Add a method to `OfflineEnvironment` to directly combine it with another.

Open Points:
- The naming is not consitent yet. The method is called `extend` the environment `Combine`. I'm open to name suggestions here
- The same functionality could be useful for `IncrementalEnvironments`. Should we extend this implementation to cover them as well or do we define a seperate environment for that task?

### DIFF
--- a/src/flowcean/core/environment/__init__.py
+++ b/src/flowcean/core/environment/__init__.py
@@ -1,6 +1,7 @@
 __all__ = [
     "ActiveEnvironment",
     "Environment",
+    "CombineEnvironment",
     "IncrementalEnvironment",
     "NotLoadedError",
     "OfflineEnvironment",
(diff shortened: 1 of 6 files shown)


schmidma:
What do you think of something like `horizontal stack` or `join`

ChristianWieck:
`horizontal stack` is technically correct, but doesn't sound "sexy". `join` is fine, but can easily be confused with `stack`

schmidma:
How do we resolve this discussion? :smile:

ChristianWieck:
Hold a vote? @newLabAspect and @mknitt what do you think? Which name is a good fit?

newLabAspect:
I prefer being technically sound over being sexy. Thus, I would rather vote for `horizontal stack`. Nevertheless, in that case we should probably re-name `stack` to `vertical stack`. To have it shorter, another option would be `vstack` and `hstack`?

PR #116, opened 2024-08-01
Add One hot transform

### LINKED ISSUE #63: One-Hot transform
Depending on the learner, it may make sense not to pass integers directly to the learner as such, but to convert them into a series of binary flags.
Example (lower limit = 1, upper limit = 5, both given(?) to the transform or fitted):

- `1 = [1, 0, 0, 0, 0]`
- `2 = [0, 1, 0, 0, 0]`
- ...
- `5 = [0, 0, 0, 0, 1]`

(Gitlab Issue 98, created by @ChristianWieck )

ChristianWieck:
This PR:
- Adds a `OneHot` transform.

Closes #63.


src/flowcean/transforms/one_hot.py
```diff
@@ -0,0 +1,71 @@
...
+        Args:
+            features: The features to transform.
+        """
+        self.features = features
+
+    @override
+    def transform(self, data: pl.DataFrame) -> pl.DataFrame:
+        for feature in self.features:
+            # Check if the feature is an integer feature
+            if not data.schema[feature].is_integer():
```
schmidma, line 52: Why only integer? Isn't this applicable to all categorical values?

ChristianWieck: It is. But we have no way in the framework to handle categorical values besides ints. We could - in theory - use `unique()` to get a set of all present values and then transform based on those... But I'm not sure I like the idea.

What I just notice: We should convert this transform into a `Supervised Learner` like `Standardize`. Otherwise results may not be consistent when calling the transform for different data frames.

schmidma: I am not sure whether this works... This would add additional columns/features to the dataframe while iterating the environment?! Does that work? The learners would have to cope with changing feature names...

For example, if the first row in an incremental environment only shows a single value, then the one hot encoding is useless, and there still would be a single feature, but as soon as the second categorical variant arrives, the one-hot encoding would add an additional feature...

schmidma: But exactly the described situation does not work for an incremental environment. If this is fitted incrementally, the input features change while training....

schmidma: I think this might be the only, or at least the best option I can currently think of... Let the user specify the categorical variants of each feature one wants to encode

ChristianWieck: It's the same problem as with `Standardize`. If we adjust to little data we might get unwanted results.
But for `OneHot` I would add the categories as an optional argument. If you have enough data / an offline environment, automatically determining the categories feels like a worthwhile addition.

src/flowcean/transforms/one_hot.py
```diff
@@ -0,0 +1,145 @@
...
+        self,
+        feature_categories: dict[str, list[Any]],
+    ) -> None:
+        """Initializes the One-Hot transform.
+
+        Args:
+            feature_categories: Dictionary of features and a list of
+                categorical values to encode for each.
+        """
+        self.feature_categrorie_mapping = {
```
mknitt, line 52: typo

src/flowcean/transforms/one_hot.py
```diff
@@ -0,0 +1,145 @@
...
+            category_mappings,
+        ) in self.feature_categrorie_mapping.items():
+            data = data.with_columns(
+                [
+                    pl.col(feature).eq(value).cast(pl.Int64).alias(name)
+                    for name, value in category_mappings.items()
+                ]
+            ).drop(feature)
+
+            # Check if any of the values is not present as a category
```
mknitt, line 94: comment not needed in my opinion

mknitt (approved):

PR #117, opened 2024-08-01
Add One-Cold transform

ChristianWieck:
This PR:

- Adds a `OneCold` transform.

### DIFF
--- a/src/flowcean/transforms/one_cold.py
+++ b/src/flowcean/transforms/one_cold.py
@@ -47,6 +47,16 @@ class OneCold(Transform):
      transform. In that case if an unknown value is found which does not belong
      to any category, a NoMatchingCategoryError is thrown. This however has an
      impact on the performance and will slow down the transform.
+
+     If you want to enable this check, create the transform as follows:
+     ```python
+        transform = OneCold(
+            feature_categories={
+                "feature": [0, 1, 2, 5]
+            },
+            check_for_missing_categories=True
+        )
+    ```
     """
 
     feature_category_mapping: dict[str, dict[str, Any]]
--- a/src/flowcean/transforms/one_cold.py
+++ b/src/flowcean/transforms/one_cold.py
@@ -41,7 +41,7 @@ class OneCold(Transform):
      1        | 1         | 1         | 0
 
      In the default configuration missing categories are ignored.
-     There respective entries will all be one. If you however want to enforce
+     Their respective entries will all be one. If you however want to enforce
      that each data entry belongs to a certain category, you can set the
      check_for_missing_categories flag to true when constructing a One-Cold
      transform. In that case if an unknown value is found which does not belong
--- a/src/flowcean/transforms/one_cold.py
+++ b/src/flowcean/transforms/one_cold.py
@@ -39,6 +39,14 @@ class OneCold(Transform):
      1        | 1         | 0         | 1
      1        | 0         | 1         | 1
      1        | 1         | 1         | 0
+
+     In the default configuration missing categories are ignored.
+     There respective entries will all be one. If you however want to enforce
+     that each data entry belongs to a certain category, you can set the
+     check_for_missing_categories flag to true when constructing a One-Cold
+     transform. In that case if an unknown value is found which does not belong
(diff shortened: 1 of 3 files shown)


schmidma (dismissed):
I'd suggest to reuse the implementation of the one-hot encoding and just take the inverse of that... What do you think?

ChristianWieck:
I decided to implement the `OneCold` transform by copying most of the `OneHot` transform. Just applying a `OneHot` transform and inverting it unfortunately isn't that easy. I could implement a shared base that is used by both transforms, but tbh I doubt it's worth it for two transforms.

src/flowcean/transforms/one_cold.py
```diff
@@ -0,0 +1,153 @@
...
+        """
+        if len(self.feature_category_mapping) == 0:
+            raise NoCategoriesError
+        for (
+            feature,
+            category_mappings,
+        ) in self.feature_category_mapping.items():
+            data = data.with_columns(
+                [
+                    pl.col(feature).ne(value).cast(pl.Int64).alias(name)
```
newLabAspect, line 95: As the polars expressions here and in lines 100-107 are quite hard to grasp, I am wondering whether it's worth to add some inline comments or extend the doc-string?

src/flowcean/transforms/one_cold.py
```diff
@@ -0,0 +1,153 @@
...
+        self,
+        feature_categories: dict[str, list[Any]],
+        *,
+        check_for_missing_categories: bool = False,
+    ) -> None:
+        """Initializes the One-Hot transform.
+
+        Args:
+            feature_categories: Dictionary of features and a list of
+                categorical values to encode for each.
```
newLabAspect, line 57: Provide an example for the dictionary? (e.g., with respect to the example above)

src/flowcean/transforms/one_cold.py
```diff
@@ -0,0 +1,153 @@
...
+            features: Name of the features for which the one hot transformation
+                will determine the categories.
+            check_for_missing_categories: If set to true, a check is performed
+                to see if all values belong to a category. If an unknown value
+                is found which does not belong to any category, a
+                NoMatchingCategoryError is thrown. To perform this check, the
+                dataframe must be materialised, resulting in a potential
+                performance decrease. Therefore it defaults to false.
+        """
+        # Derive categories from the data frame
```
newLabAspect, line 135: If we decide not to put inline comments, we should probably remove this one as well.

newLabAspect (commented):
The implementation looks fine to me. Just had a hard time to understand what's going on. Thus, some suggestions on how to improve this for users.

src/flowcean/transforms/one_cold.py
ChristianWieck: Do you mean the flag or the comment? I think it's still relevant for the user to have the choice - they don't need to understand how it works under the hood as long as it does what they want :D

src/flowcean/transforms/one_cold.py
ChristianWieck: Added a code example to the "main-docstring".

src/flowcean/transforms/one_cold.py
```diff
@@ -39,6 +39,14 @@ class OneCold(Transform):
      1        | 1         | 0         | 1
      1        | 0         | 1         | 1
      1        | 1         | 1         | 0
+
+     In the default configuration missing categories are ignored.
+     There respective entries will all be one. If you however want to enforce
```
newLabAspect, line 44: There -> Their

newLabAspect (commented):
Found a minor typo. Maybe we can fix that directly ;)

src/flowcean/transforms/one_cold.py
newLabAspect: I just meant the comment ;)

src/flowcean/transforms/one_cold.py
ChristianWieck: Should be fixed.

newLabAspect (approved):
lgtm now

ChristianWieck:
@schmidma what do you think?

PR #122, opened 2024-08-01
Add Resample transform

ChristianWieck:
This PR:
- Adds a new implementation of a `Resample` transform to change the sampling rate of time series features.

Probably superseeds #23 as this uses the new time series format.

The implementation could possible still be improved quite a bit. As I use interpolation methods from scipy there is probably no way around moving the data from polars to python and back. @schmidma maybe you have another clever idea how to optimize it further / make it more polars.

### DIFF
--- a/src/flowcean/transforms/resample.py
+++ b/src/flowcean/transforms/resample.py
@@ -1,9 +1,8 @@
 import logging
 import math
-from typing import Any, cast, override
+from typing import cast, override
 
 import numpy as np
-import numpy.typing as npt
 import polars as pl
 from scipy.interpolate import CubicSpline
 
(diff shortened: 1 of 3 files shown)


src/flowcean/transforms/resample.py
```diff
@@ -0,0 +1,104 @@
...
+
+
+class Resample(Transform):
+    """Resample time series features to a given sampling rate."""
+
+    def __init__(
+        self,
+        sampling_rate: float | dict[str, float],
+        *,
+        use_cubic_interpolation: bool = False,
```
schmidma, line 22: Maybe do not use `bool` but rather a string selecting the different methods.

schmidma (changes requested):
one minor thing... For now I also do not have any more fancy ideas on how to do that, unless we extend polars and do it in Rust

src/flowcean/transforms/resample.py
ChristianWieck: Sure, I changed it to a string type. Since we don't use enums anywhere else, I didn't feel like introducing them here.

ChristianWieck:
> one minor thing... For now I also do not have any more fancy ideas on how to do that, unless we extend polars and do it in Rust

Then we'll have to deal with the `map_elements` for now. Once we see how the performance impact is for real data we can still re-evaluate imho.

schmidma (approved):

PR #126, opened 2024-08-05
Add workaround for json loader

ChristianWieck:
This PR:
- Adds a check for the `JsonDataloader` to infer the right data schema and not implicitly repeat data

Without this PR a json file of the format
```json
{
  "feature_a": 5,
  "feature_b": [1,2]
}
```
would be loaded as

feature_a | feature_b
----------|-----------
 5            | 1
 5            | 2

which is most likely not what the user intended.
With this PR, the loaded data will be

feature_a | feature_b
----------|-----------
 5            | [1, 2]

to achieve the previous results the input json should be

```json
{
  "feature_a": [5,5],
  "feature_b": [1,2]
}
```

### DIFF
--- a/src/flowcean/environments/json.py
+++ b/src/flowcean/environments/json.py
@@ -19,7 +19,15 @@ class JsonDataLoader(OfflineEnvironment):
     def load(self) -> Self:
         with self.path.open() as file:
             json_content = json.load(file)
-        self.data = pl.DataFrame(json_content)
+
+        # Check if any of the entries in the dict is *not* a list and treat the
+        # whole dict as a single entry in that case
+        if isinstance(json_content, dict) and any(
+            not isinstance(value, list) for value in json_content.values()
+        ):
+            self.data = pl.DataFrame([json_content])
+        else:
+            self.data = pl.DataFrame(json_content)
         return self
 
     @override


schmidma:
I think this change is not necessary: in the current main, you can have the same results if you put the following dict/json:

```
{
  "a": [5],
  "b": [[1, 2]],
}
```

In my eyes, this even shows the schema more clearly. But this might be just my opinion.

What do you think?

ChristianWieck:
I agree that this is not optimal. But with the current implementation polars will automaticlly add rows to the resulting dataframe even if there isn't enough data in all features (and just repeat the values). If you have lists of different length polars will throw an error.

What this basiclly does is allow the user to provided higher-dimensional data (say lists) for a single sample and avoiding this behaviour (which I guess is not desired - atleast I was supprised when I found out that the ALP data was loaded as five samples and not one).

schmidma:
I'm not sure whether this is a misunderstanding. At least, I did not get how your reply relates to my suggestion...

What I were suggesting was to do nothing. The current implementation allows the same shape of a DataFrame as you want to have (with a list as a single element). So I do not see the reason to change anything. Instead of changing the code to fit your parsing structure, you might also just change your jsons.

So I am arguing to keep the code as is, as it has more features (the automatic adding of rows for single value features). Your proposed change removes this feature

ChristianWieck:
That's right. I just feel that it's more one of thoses features nobody has asked for and which can (and actually already has for me) cause confusion and works incosistently.

We can totally keep the loader as is - in that case I only need to update my ALP data again to work with it

schmidma:
Yeah, that is a valid point. We wanted to have the framework understand various input formats and cope with that in the transforms. But I think that is something you can already do. So, load the json as is, with the code as is, and fix the format afterwards using transforms.

My main motivation is to keep the interface to polars as lean as possible and do not change any API. Otherwise this also might be confusing if you already know polars ...

ChristianWieck:
I would argue that most of the users probably don't have any polars expirence. But I'm okay with keeping the API as-is. I'll update the ALP data to work with the current implementation.

PR #129, opened 2024-08-05
Add Time series filter

### LINKED ISSUE #34: Time Series - Lowpass Filter
(Gitlab Issue 114, created by @ChristianWieck)

### LINKED ISSUE #38: TimeSeries - Filter/Convolution transform
(Gitlab Issue 108, created by @ChristianWieck

ChristianWieck:
This PR:
- Adds a `Filter` transform to apply high- and lowpass (butterworth) filters to time series data.

We might want to add a bandstop / bandpass as well, what do you think?
Also, a new filter is created for each sample. This could be reworked to create only one filter per feature, but that would require knowing the sampling frequency a priori. I choose convenience over performance here.

This transformation uses `map_elements` and can be quite slow. However, since I use scipy filter functions, there is no real way around this.

Closes #34, Closes #38.

### DIFF
--- a/src/flowcean/transforms/__init__.py
+++ b/src/flowcean/transforms/__init__.py
@@ -1,5 +1,7 @@
 __all__ = [
     "Explode",
+    "Filter",
+    "FilterType",
     "FeatureLengthVaryError",
     "Flatten",
     "NoTimeSeriesFeatureError",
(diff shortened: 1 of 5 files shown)


mknitt (approved):
lgtm

PR #133, opened 2024-08-06
tqdm progress bar in env.collect()

schmidma:
Adds a progress bar to the collecting of an incremental environment.


schmidma:
I added `tqdm` as a dependency

ChristianWieck (approved):

PR #140, opened 2024-08-07
Add Time series window transform

### LINKED ISSUE #139: Create TimeWindow transform
Create a transform to limit time series to certain time windows.

ChristianWieck:
This PR:
- Adds a `TimeWindow` transform to limit time series features to a certain time window. Data outside the window is dropped.

Closes #139.


src/flowcean/transforms/time_window.py
```diff
@@ -0,0 +1,64 @@
...
+        ):
+            time_expression = (
+                pl.element().struct.field("time").cast(pl.Float64)
+            )
+
+            data = data.with_columns(
+                pl.col(feature).list.eval(
+                    pl.element().filter(
+                        time_expression.ge(self.t_start).and_(
+                            time_expression.le(self.t_end)
```
schmidma, line 59: Can you use `>=` or `<=`? Is this supported?

ChristianWieck: I don't think so. The polars examples always use `eq`, `ge` and `le` (and lable the resulting columns with `<=`). So I guess we're "stuck" with `ge` and `le`.

schmidma (approved):

PR #141, opened 2024-08-07
Support lazy frames in transforms

ChristianWieck:
This PR:
- Modifies existing transforms to expect lazy dataframes. This can - potentially - increase performance.
- Add a `ToLazy` and `Collect` transform which can be used to transform data from/to lazy representations.

This PR is part of greater effort to allow the usage of lazy frames within flowcean.

Waiting for #160.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -15,7 +15,7 @@ authors = [
 
 dependencies = [
   "numpy<2",
-  "polars",
+  "polars>=1.4",
   "ruamel.yaml",
   "typing-extensions",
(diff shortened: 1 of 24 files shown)


ChristianWieck:
I'm not sure why the CI/CD fails here. Running on my machine with polars `1.4.1` `hatch --verbose run types:check` yield no error. Is it possible that the CI/CD uses an old polars version @schmidma ?

ChristianWieck:
Outdated after #160.

PR #144, opened 2024-08-07
Make environment builder/loaders return lazyframes where possible

ChristianWieck:
This PR:
- Allow environments to return `LazyFrames`
- Adjust existing environments to make use of this feature where possible.

Depends and requires #141.


ChristianWieck:
Outdated after #160.

PR #146, opened 2024-08-08
Add `write_parquet` method to offline environment

ChristianWieck:
This PR:
- Adds a `write_parquet` to the offline environment. This allows the user to store data after loading / transforming it and can save time when making changes to the learners / metrics without changing the loading part. Does also support the time series representation.

This is "just" a wrapper around polars functionality - however I would argue that it's easier for a user to have a "flowcean-method" then having to dive into the polars docs. Also this makes sure the writing shows up in our documentation. If you need more options you can still use them by directly using the data frame.

### DIFF
--- a/src/flowcean/core/environment/offline.py
+++ b/src/flowcean/core/environment/offline.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 from abc import abstractmethod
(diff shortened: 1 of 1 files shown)


newLabAspect (approved):
lgtm

PR #147, opened 2024-08-08
Pack ALP data into single files

ChristianWieck:
This PR:
- Repacks the ALP data into single `parquet` files and adjust the example accordingly. This makes working with the data much easier. The helper script to convert the original alp data into a single file is included as well.

### DIFF
--- a/examples/automatic_lashing_platform/.gitignore
+++ b/examples/automatic_lashing_platform/.gitignore
@@ -1,3 +1,0 @@
-models/
-/data
-/real_data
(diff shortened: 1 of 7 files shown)


schmidma (changes requested):
If I understand this correctly, we still need the many individual files, otherwise the `preload_alp_data.py` does not work... We did you remove them?

ChristianWieck:
That's right. `preload_alp_data.py` was used to convert the many individual files into the single `.parquet`. I keept it in the repo to be able to convert new data from the "many-files" format into the "singel-file", so it's more a helper script than part of the example. I've removed the individual files as they have no advantage over the packed file. The packed file is smaller, can be downloaded faster and loaded into flowcean faster while still having the same properties as the loaded single files.

schmidma (approved):

PR #148, opened 2024-08-13
Update landing page

schmidma:
Provides some advertising text on the landing page. This has to be updated as soon as we have references to a paper we want to cite. Additionally, it claims to be installable via `pip install flowcean` which is currently not possible.


ChristianWieck:
Lgtm. Regarding the PyPI release: We should probably define a state when flowcean is "good enough" for a first alpha / beta release. I'll throw a discussion issue on the tracker.

PR #158, opened 2024-08-24
Adapt coffee machine to time series

newLabAspect:
- re-define protobuf and re-generate files
- refactor coffee machine example, grpc automata learner and Java code for time-series
- rename grpc learner to grpc passive automata learner and move proto files to its definiton
- add evaluation transforms to offline strategies
- add unnest transform


src/flowcean/learners/grpc_passive_automata/learner.py
```diff
@@ -68,19 +71,23 @@ def __init__(
...
+            remove=False,
         )
         if isinstance(container, Container):
             self._docker_container = container
         else:
             message = "did not receive a container"
             raise TypeError(message)
         self._docker_container.reload()
-        ports = self._docker_container.ports
+        client = docker.APIClient(base_url="unix://var/run/docker.sock")
```
ChristianWieck, line 86: Does this work on windows as well?

src/flowcean/learners/grpc_passive_automata/proto/protoc-gen-grpc-java:Zone.Identifier
ChristianWieck, line 1: This looks like a windows ads artefact. Can you delete it?

src/flowcean/strategies/offline.py
```diff
@@ -64,6 +64,7 @@ def evaluate_offline(
     inputs: list[str],
     outputs: list[str],
     metrics: list[OfflineMetric],
+    eval_transform: Transform | None = None,
```
ChristianWieck, line 67: What is the purpose of this transform? How is it different from `input_transform` in the current `main`?

src/flowcean/transforms/unnest.py
ChristianWieck, line 1: Can we move this transform into it's own PR? It's also missing some documentation and a testcase...

ChristianWieck (commented):
As far as I can tell, this changes the gRPC interface to be exclusively "time series" based. While I think this is a good thing to get the automaton learner working, we should agree as a team to take this step.

src/flowcean/learners/grpc_passive_automata/learner.py
newLabAspect: I removed the explicit base-url. By this, the API client seems to internally check the system type. Anyways, I did not explicitly test this on windows.

src/flowcean/learners/grpc_passive_automata/proto/protoc-gen-grpc-java:Zone.Identifier
newLabAspect: Done

src/flowcean/transforms/unnest.py
newLabAspect: I moved this to #244, added a test case and adapted the docu there. I leave the transform here as well as this branch would otherwise not be exectuable. Nevertheless, this meand that this PR now depends on #244 .

src/flowcean/strategies/offline.py
ChristianWieck: But isn't this equal to the `output_transform` of a `ModelWithTransforms`?

src/flowcean/transforms/unnest.py
ChristianWieck: 👍🏻

src/flowcean/learners/grpc_passive_automata/learner.py
ChristianWieck: If I find a moment I'll test it on a windows machine. I have doubts this will work on windows as well - if that's the case we can add a "plattform-guard" to ensure the gRPC learner can (for now) only be used with *nix systems)

src/flowcean/strategies/offline.py
newLabAspect: Yes, I think this is the same. So far, these output transforms are meant to be the inverse of an output transform used during learning. For the coffee machine, the transform is only applied for the evaluation and not during learning. I am wondering whether this is something that we would like to separate.

In other words, the output transform of a model is something that needs to be applied to map the output back to match with the real-world. Here, the transform is needed to apply the metrics. Thus, to me it makes more sense to attach this transform to the evaluation strategy, but this was also done at a point where we didn't have the output transforms, yet.

Technically, though, it should work here to attach the transform to the model. Is this something conceptual to discuss or should I just move the transforms to the model?

src/flowcean/learners/grpc_passive_automata/learner.py
ChristianWieck: Unfortunately, I can't test the code - I'm still missing access to the current DVD storage (I have the credentials, but the required VPN access is difficult for me).

ChristianWieck: That's great new :)

src/flowcean/strategies/offline.py
ChristianWieck: I would suggest moving the transformation to the model, but if you want to discuss it with the team, we can certainly do that.

newLabAspect: Ok, I moved it to the model, now. I am not sure whether it is used in the intended way... I will open a discussion issue to see whether there are any further thoughts on that.

ChristianWieck (approved):

PR #160, opened 2024-09-24
Generalized lego bricks

schmidma:
This PR refactors the underlying structure of environments into composable blocks of _observing_, _stepping_, and _acting_ in an environment.


ChristianWieck (approved):
Looks good to me.

PR #165, opened 2024-10-24
Pin polars version to 1.9

### LINKED ISSUE #164: Timeseries handling Bug with Polars > 1.9
There seems to be a bug when handling our time series with polars > 1.9.

While this code works with `polars=1.9` it does not with any later version.
```python
data = data.with_columns(
                [
                    pl.col(feature)
                    .list.eval(pl.element().struct.field("value"))
                    .list.get(i)
                    .alias(f"{feature}_{i}")
                    for i in range(n)
                ]
```
The error seems to be that the field `value` isn't found. I'm not sure if this is a problem with polars or changed behaviour.
A quick fix would be to pin `polars` for now at `1.9` until we find a more permanent solution.

Edit:
This seems to be the issue: https://github.com/pola-rs/polars/issues/19345. `pl.element()` is currently broken.

ChristianWieck:
This PR:
- Pins `polars` at `1.9` until the real problem with the time series handling is found.

(Temporarily) resolves #164

### DIFF
--- a/src/flowcean/transforms/flatten.py
+++ b/src/flowcean/transforms/flatten.py
@@ -63,11 +63,7 @@ class Flatten(Transform):
 
             # Figure out how "long" the feature is and how many new columns
             # need to be added
-            row_lengths = data.select(
-                pl.col(feature)
-                .list.eval(pl.first().struct.field("value"))
-                .list.len()
-            ).unique()
+            row_lengths = data.select(pl.col(feature).list.len()).unique()
 
             # Check if all rows have the same length
             if row_lengths.count().item(0, 0) > 1:
@@ -79,7 +75,7 @@ class Flatten(Transform):
             data = data.with_columns(
                 [
                     pl.col(feature)
-                    .list.eval(pl.first().struct.field("value"))
+                    .list.eval(pl.element().struct.field("value"))
                     .list.get(i)
                     .alias(f"{feature}_{i}")
                     for i in range(n)
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -20,7 +20,7 @@ authors = [
 
 dependencies = [
   "numpy<2",
-  "polars",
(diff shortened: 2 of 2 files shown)


src/flowcean/transforms/flatten.py
```diff
@@ -63,11 +63,7 @@ def transform(self, data: pl.DataFrame) -> pl.DataFrame:
 
             # Figure out how "long" the feature is and how many new columns
             # need to be added
-            row_lengths = data.select(
-                pl.col(feature)
-                .list.eval(pl.first().struct.field("value"))
-                .list.len()
-            ).unique()
```
schmidma, line 70: Was this change intended?

src/flowcean/transforms/flatten.py
```diff
@@ -79,7 +75,7 @@ def transform(self, data: pl.DataFrame) -> pl.DataFrame:
             data = data.with_columns(
                 [
                     pl.col(feature)
-                    .list.eval(pl.first().struct.field("value"))
```
schmidma, line 82: same here?

src/flowcean/transforms/flatten.py
ChristianWieck: That's a leftover artefact from trying to figure out if the error was in our code or in the polars. In the process, I realised that the check was overly complicated (we don't really need to unpack the strucst to know how many there are). So the change is "good", it just doesn't necessarily belong in this PR.

src/flowcean/transforms/flatten.py
ChristianWieck: Same as above, a leftover from debugging. Polars recommends using `pl.element()` to get the current element in an `eval` expression. However, `pl.first()` works just as well - I just wanted to see if that's the breaking bug.

schmidma (approved):
lgtm

PR #166, opened 2024-10-28
Unpin polars and limit version to latest one with the bugfix

ChristianWieck:
This PR:
- Does what the title says. Polars since version `1.12` has the bugfix for list `evals`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -20,7 +20,7 @@ authors = [
 
 dependencies = [
   "numpy<2",
-  "polars==1.9",
+  "polars>=1.12",
   "ruamel.yaml",
   "typing-extensions",
   "tqdm",


ChristianWieck:
@schmidma the tests for the ode environment fail, although I haven't touched them... I guess there is some API change in Polars `1.10` onwards that causes this?

schmidma:
> @schmidma the tests for the ode environment fail, although I haven't touched them... I guess there is some API change in Polars `1.10` onwards that causes this?

Yes, looks like it. But I guess, we have to fix it, to actually get this version bump merged...

ChristianWieck:
The changes from #160 seem to have fixed the issue. Tests are passing now.

schmidma (approved):

PR #167, opened 2024-10-28
uv for project management

schmidma:
- Use uv (https://docs.astral.sh/uv/) for project management.
- introduce additional checks via `pre-commit` (https://pre-commit.com/) ... and fixes them
- adjust and improve documentation accordingly

PR #171, opened 2024-11-14
pin python version to 3.12

COOOkies4U:
This should be correct if any version of 3.12 is acceptable.

Otherwise, something like that is possible: requires-python = ">=3.12.3, <3.13"

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -7,7 +7,7 @@ name = "flowcean"
 dynamic = ["version"]
 description = "Automatic generation of models for cyber-physical systems."
 readme = "README.md"
-requires-python = ">=3.12"
+requires-python = "==3.12.*"
 classifiers = [
     "Programming Language :: Python :: 3",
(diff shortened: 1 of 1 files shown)


schmidma:
Thanks for this PR, this is a useful addition ;)

I think we can use the _compatible release_ specifier (https://packaging.python.org/en/latest/specifications/version-specifiers/#compatible-release), can we?

COOOkies4U:
That's right, we can take that one. But do we need it? If so, what is the minimum version required?

COOOkies4U:
That's true, but only if we need versions greater than or equal to 3.12.7, right? Isn't 3.12.3 sufficient?

schmidma:
@ChristianWieck was extremely fast in merging #167 and it is superseding this PR... so fixed with #167

COOOkies4U:
Also very nice. I'm excited to see how UV is used.

PR #172, opened 2024-11-14
Migrate dataframes to lazy

ChristianWieck:
This PR:
- Changes the datatype of (almost) all `DataFrame` to `LazyFrame`.


src/flowcean/core/model.py
```diff
@@ -62,8 +62,8 @@ def predict(
         self,
         input_features: pl.DataFrame,
```
schmidma, line 63: Why not having these a LazyFrame as well?

src/flowcean/core/environment/offline.py
```diff
@@ -73,7 +73,9 @@ def write_parquet(self, path: Path | str) -> None:
         Args:
             path: Path to the parquet file where the data is written.
         """
-        self.observe().write_parquet(Path(path).with_suffix(".parquet"))
+        self.observe().collect(streaming=True).write_parquet(
```
schmidma, line 76: IIRC streaming is considered an unstable feature. Should we already use it?

ChristianWieck: It is - but if you want to use large datasets, streaming is key. My idea was to have a `flowcean.use_streaming` field that can be toggled by the user. What do you think about this?

src/flowcean/core/model.py
ChristianWieck: I'm not sure.. I believe the reason was that I wanted to avoid to `Collect` in every model implementation.

src/flowcean/core/environment/offline.py
schmidma: Did you ever encounter issues using this "unstable" feature? Otherwise, we just go with it, and introduce such a toggle later if issues arise

ChristianWieck: Nope, worked fine for me. I guess the problems arise if your pipeline includes some more exotic stuff. I would totally agree with a "just-roll-with-it-and-fix-stuff-if-it-breaks" approach here 😄

schmidma (approved):

PR #179, opened 2024-11-19
Add Filter transform

### LINKED ISSUE #150: Filter transform
Introduce a `Filter` transform which allows for filtering an environment. This is easy to do with polars, the question is how we want to expose the functionality to users.

ChristianWieck:
This PR:
- Add a new `Filter` transform to filter environments based on one or multiple expressions. The code is not super nice and uses `sympy.parse_expr` under the hood, however the expressions are translated to polars and should be as fast as native "polars expressions". `Filter` now supports more complex expressions using `And`, `Or` and `Not` types to combine expressions.

~~This approach only allows for an `and` or `or` filter and not for combined expressions. Alternatives could be to~~
- ~~Split `Filter` into `FilterAnd` and `FilterOr` to get rid of the `filter_method` argument~~
- ~~Introduce `And` and `Or` types to be used as arguments to `Filter` like this~~
  ```python
    transform = Filter(And("x > 2", Or("y < z", "w != 7")))
  ```

Closes #150.


newLabAspect:
The syntax looks quite intuitive to me, now if we add 1-2 nice examples to the doc string :)

ChristianWieck:
> The syntax looks quite intuitive to me, now if we add 1-2 nice examples to the doc string :)

There shoud now be a few examples in the docstring 😄

newLabAspect (approved):
lgtm

PR #184, opened 2024-12-13
Batch transform

mknitt:
Based on your sketch @ChristianWieck, I wrote this transform, and added some tests.


mknitt:
Closing this for now, as I think we do not necessarily need that functionality for now.

PR #163, opened 2024-10-15
Particle clouds

mknitt:
- Removes dependency for the rosbags-dataframe package (which includes pandas)
- Adds support for loading multidimensional and nested ROS data such as particle clouds 
- Demonstration of the new functionality in `/examples/ros_offline/run.py `

PR #187, opened 2024-12-19
Scan and map transforms

mknitt:
Depends on #163 

This adds functions to extract features from a laser scan and an occupancy grid map. 

Not sure if I should create transforms for that, or if the necessary functions should just remain in the here added python files in the example folder. @ChristianWieck What do you think?


pyproject.toml
```diff
@@ -21,9 +21,13 @@ dependencies = [
     "grpcio>=1.67.0",
     "joblib>=1.4.2",
     "lightning>=2.4.0",
+    "matplotlib==3.10.0",
     "numpy>=1.26.4",
+    "opencv-python>=4.10.0.84",
     "polars>=1.9.0",
     "protobuf>=5.28.2",
+    "PyQt5-Qt5==5.15.15",
+    "PyQt5_sip==12.16.1",
```
ChristianWieck, line 30: Can you move the dependencies into your example? We don't require them for flowcean but "just" for the example. See https://github.com/flowcean/flowcean/blob/9ea9664e34464c998fda77c79533dfdaf89c34ee/examples/coffee_machine/run.py#L1C1-L6C6.

ChristianWieck (changes requested):
Before we merge this PR, we should try to separate the "new" changes from the tweaking of existing code. For example, the rosbag environment changes a lot and I'm not in a good position to review those changes. I would suggest you open a separate PR for them.

Regarding the helper functions: If their use is limited to your example, I'm fine with keeping them as loose files in the examples folder. However, if we can generalise them to arbitrary images (e.g. line detection might be useful for others too?), we should make a transformation out of them. This is closely related to the whole "higher dimensional data" thing that we haven't discussed yet...

pyproject.toml
mknitt: I moved the dependencies into the example.

mknitt: The current implementation of the rosbag environment does not support scan and map data to be loaded. So, this PR requires the rosbag environment to be changed. That's why the changes to the rosbag environment are included in this PR. Who would you suggest for reviewing the changes to the rosbag environment? I propose that we keep the changes related to the rosbag environment in this PR for the sake of simplicity. 

Regarding the helper functions: I propose we keep them in this way for now and do the generalization to arbitrary images once another use case requires line detection. What do you think?

examples/robot_localization_failure/utils.py
```diff
@@ -0,0 +1,214 @@
+import math
+
+import cv2
+import matplotlib.pyplot as plt
```
ChristianWieck, line 4: These dependencies are'n listed anywhere, are they? Can you add them to the example "dependency comment"?

examples/robot_localization_failure/utils.py
```diff
@@ -0,0 +1,214 @@
...
+    t2 = 2.0 * (w * y - z * x)
+    t2 = max(-1.0, min(1.0, t2))  # Clamp to handle numerical errors
+    pitch = math.asin(t2)
+
+    # Yaw (z-axis rotation)
+    t3 = 2.0 * (w * z + x * y)
+    t4 = 1.0 - 2.0 * (y * y + z * z)
+    yaw = math.atan2(t3, t4)
+
+    return roll, pitch, yaw
```
ChristianWieck, line 35: This probably works, but why aren't you using https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.transform.Rotation.html? `scipy` is a dependency of flowcean anyway.

ChristianWieck (commented):
As this is only relevant to your example, I'm fine with merging as-is.

examples/robot_localization_failure/utils.py
mknitt: Thanks, good tip! I am using scipy for this now.

examples/robot_localization_failure/utils.py
mknitt: I moved them into the dev-dependencies.

mknitt: I had to move them back to the dependencies because the CI checks would not run through, although the dependencies are in the example dependency comment.

mknitt:
> As this is only relevant to your example, I'm fine with merging as-is.

Since this comment, I moved the scripts from the example folder to a single transform, added some more documentation, and improved the execution speed.

examples/robot_localization_failure/utils.py
ChristianWieck: Hmm that's strange. @schmidma do you have an idea why this might happen? I would rather not have open-cv as a flowcean dependency...

mknitt: As we discussed, the dependencies are required because they are needed for the transform itself and not the run.py of the example. Previously, I had the plotting in a utils.py in the example folder, but I think it is nice for developers to tune the transform and getting some visual feedback via the plots. As long as this does not break anything else, I vote for just accepting to have the dependencies for now. 

@ChristianWieck What do you think?

ChristianWieck: Okay for me. I guess with #216 we can modularize our dependencies more and allow the user to install only those they need.

src/flowcean/transforms/scan_map.py
```diff
@@ -0,0 +1,409 @@
...
+    data in conjunction with an occupancy grid map. By combining pose
+    information from the sensor and scan data, this class computes scan point
+    coordinates in the map frame and calculates distances between scan points
+    and detected lines in the occupancy grid.
+    """
+
+    def __init__(
+        self,
+        scan_topic: str = "/scan",
+        sensor_pose_topic: str = "/amcl_pose",
```
ChristianWieck, line 27: Do we need this strong ROS coupling here? I believe a flowcean transform shoudn't assume that the data comes from ros but rather work on "arbitrary pointcloud"?

src/flowcean/transforms/scan_map.py
```diff
@@ -0,0 +1,409 @@
...
+            orientation_z = pose_data["pose.pose.orientation.z"]
+            orientation_w = pose_data["pose.pose.orientation.w"]
+            theta = Rotation.from_quat(
+                [
+                    orientation_x,
+                    orientation_y,
+                    orientation_z,
+                    orientation_w,
+                ],
+            ).as_euler("xyz", degrees=False)[2]
```
ChristianWieck, line 75: Just an implementation detail: Do you really want to use "xyz"? Whenever I wanted to convert a quaternion into a yaw, "zyx" worked more reliably.

src/flowcean/transforms/scan_map.py
ChristianWieck, line 1: I think this transform is highly specialised for your use case and can't really be used outside of it. While that's not a problem in itself, I would generally prefer to move the transform to the "example" folder, since it can't be used by a broad audience outside of that folder. This could work as an example of "how to write and use custom transforms".

src/flowcean/transforms/scan_map.py
mknitt: Most people that have LiDAR scan data and occlusion maps use ROS. Also, I followed your other advice and moved everything into a `custom_transforms` folder in the robot_localization_failure example. Therefore, we can keep the strong ROS coupling here.

src/flowcean/transforms/scan_map.py
mknitt: You are right, that makes more sense. I moved this transform and `ParticleCloudStats` another use case specific transform into the "example" folder.

examples/robot_localization_failure/utils.py
mknitt: Now that the transforms are moved into the example, I removed the dependencies for OpenCV and Matplotlib from the `pyproject.toml`.

pyproject.toml
ChristianWieck, line 1: With the new structure and the transform moved into a example, those dependencies should be not necessary anymore, no?

src/flowcean/transforms/scan_map.py
ChristianWieck: Actually there is a difference, because with "xyz" small roll and pitch movements will affect your calculated yaw angle, whereas with "zyx" the yaw angle is independent of such movements.

src/flowcean/transforms/scan_map.py
ChristianWieck: I'm still not completly convinced that these transforms need such strong coupling but for the sake of getting this merged, we can keep it as-is for now.

examples/robot_localization_failure/utils.py
ChristianWieck: To be honhest I'm not sure if the `dev-dependencies` are the right place for those. My understanding of these was to include tools which are required during developtment but not deployment (such as `ruff`). If we keep the transform specific to on example I would strongly vote to exclude `open-cv` and `matplotlib` from the `pyproject.toml` and only put them into the example dependencies / the readme.

ChristianWieck: Decided to leave them as regular dependencies for now. I got a plan how to fix this.

ChristianWieck (dismissed):

SM20598 (approved):

PR #189, opened 2025-01-14
Add flowcean as editable dependency to scripts

schmidma:
Currently, flowcean is installed from official released sources (pypi) when scripts are run with `uv`.
This PR adds a section telling `uv` to use flowcean from the local path instead.


ChristianWieck (approved):
Lgtm

PR #199, opened 2025-01-14
Output Transforms

schmidma:
This adds output transforms as discussed on the concept board last week.
Output transforms are applied to outputs when learning in a supervised fashion, the inverse of this transform is also attached to the model, to maintain equivalent interfaces when using the model.

### DIFF
--- a/src/flowcean/core/model.py
+++ b/src/flowcean/core/model.py
@@ -55,15 +55,23 @@ class ModelWithTransform(Model):
     """
 
     model: Model
-    transform: Transform
+    input_transform: Transform | None
+    output_transform: Transform | None
 
     @override
     def predict(
         self,
         input_features: pl.LazyFrame,
     ) -> pl.LazyFrame:
-        transformed = self.transform.apply(input_features)
-        return self.model.predict(transformed)
+        if self.input_transform is not None:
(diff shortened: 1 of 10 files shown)


src/flowcean/core/model.py
```diff
@@ -55,15 +55,23 @@ class ModelWithTransform(Model):
...
     ) -> pl.LazyFrame:
-        transformed = self.transform.apply(input_features)
-        return self.model.predict(transformed)
+        if self.input_transform is not None:
+            input_features = self.input_transform.apply(input_features)
+
+        prediction = self.model.predict(input_features)
+
+        if self.output_transform is None:
+            return prediction
```
ChristianWieck, line 72: Stylistic nitpicking, but I would try to keep the `ifs' consistent - either both `is` or both `is not`. Currently they express the exact opposite statements.

schmidma: Took some time to understand what you meant :smile: But I replaced this if with `if self.output_transform is not None:` and swapped the return paths.

ChristianWieck: 👍🏻

ChristianWieck (approved):

ChristianWieck (approved):

Issue #192, opened 2025-01-14
Tool Strategy


newLabAspect:
- like learning and evaluation strategy for tools
- find suitable naming for this type of strategy
- implements from ConceptBoard

PR #204, opened 2025-01-21
Deployment strategies

SM20598:
Deployment strategy based on #192


src/flowcean/strategies/deployment.py
```diff
@@ -0,0 +1,50 @@
...
+from flowcean.core.environment.incremental import IncrementalEnvironment
+from flowcean.core.transform import ChainedTransforms
+
+from flowcean.core.model import ModelWithTransform
+
+def deploy(
+    environment: ActiveEnvironment | IncrementalEnvironment,
+    config: list[str],
+    model: ModelWithTransform,
+    input_transforms: ChainedTransforms,
```
schmidma, line 12: This type bound probably only needs to be `Transform`
```suggestion
    input_transforms: Transform,
```

src/flowcean/strategies/deployment.py
```diff
@@ -0,0 +1,50 @@
+import polars as pl
+from flowcean.core.environment.active import ActiveEnvironment
+from flowcean.core.environment.incremental import IncrementalEnvironment
+from flowcean.core.transform import ChainedTransforms
+
+from flowcean.core.model import ModelWithTransform
+
+def deploy(
+    environment: ActiveEnvironment | IncrementalEnvironment,
+    config: list[str],
```
schmidma, line 10: This seems to be unused?

src/flowcean/strategies/deployment.py
```diff
@@ -0,0 +1,50 @@
...
+    Use a trained model
+
+    Args:
+        observation: an observation of the environments current state
+        model: the trained model
+        input_transforms: system specific transforms for model input
+        output_transforms: system specific transforms for system input
+    """
+    output = model.predict(input_transforms.apply(observation))
+    return output_transforms.apply(output)
```
schmidma, line 50: What is this function for?

src/flowcean/strategies/deployment.py
SM20598: Done

src/flowcean/strategies/deployment.py
SM20598: @mknitt and I discussed how a connection between flowcean and a robot system would look like. Our understanding is that an environment can be directly included in flowcean (e.g. a custom simulation environment) or exist externally. For example. a robot running on ROS might have a node that is calling directly applying a model trained in flowcean and provides an observation.

src/flowcean/strategies/deployment.py
SM20598: Our diagram showed a config for the environment, which is why I included a param for this function that is unused right now.

src/flowcean/strategies/deployment.py
schmidma: I'd argue, to then call `model.predict()` ... I do not see the point of this function :thinking:

src/flowcean/strategies/deployment.py
schmidma: Then, I'd remove it for now and introduce the parameter as soon as we need it

schmidma (approved):

PR #205, opened 2025-01-21
Data can be Any-thing

schmidma:
Annotate all core methods with `Data` which in turn is just an alias for `Any`

Based on #199


ChristianWieck:
With this change, is there any checking that environments, transforms and learners fit together? I think we discussed some sort of "up front" type checking. The key question imo is whether we should migrate the "anything data" first and add type checking afterwards, or add them together.

schmidma:
No there is not... This PR just drops all our requirements. We can discuss this in the team, but I personally would first drop it and add "safety" features incrementally

ChristianWieck (dismissed):

ChristianWieck (approved):

PR #210, opened 2025-01-31
Add support for Python 3.10

mknitt:

PR #211, opened 2025-01-31
ROS Package for Deploying Flowcean Models 

mknitt:
Depends on #210


mknitt:
This moved to its own repository.
https://github.com/flowcean/flowcean-ros

PR #222, opened 2025-02-14
Particle cloud image transform

manavthakkar:
Implement particle cloud image transform as mentioned in section 3.1 of this [paper](https://link.springer.com/article/10.1007/s10489-020-02157-6).


src/flowcean/polars/transforms/particle_cloud_image.py
```diff
@@ -0,0 +1,146 @@
...
+        cutting_area: float = 15.0,
+        image_pixel_size: int = 300,
+        *,
+        save_images: bool = False,
+    ) -> None:
+        """Initialize the ParticleCloudImage transform.
+
+        Args:
+            particle_cloud_feature_name: Name of the particle cloud feature.
+            cutting_area: Square region side length for cropping (in meters).
```
mknitt, line 35:
```suggestion
            cutting_area: Side length of the square region used for cropping (in meters).
```

src/flowcean/polars/transforms/particle_cloud_image.py
mknitt, line 1: this should be moved to `examples/robot_localization_failure/custom_transforms`

tests/polars/transforms/test_particle_cloud_image.py
mknitt, line 1: this should be moved to `examples/robot_localization_failure/tests`

tests/polars/transforms/test_particle_cloud_statistics.py
mknitt, line 1: this should be moved to `examples/robot_localization_failure/tests`

pyproject.toml
```diff
@@ -23,6 +23,7 @@ dependencies = [
   "joblib>=1.4.2",
   "lightning>=2.4.0",
   "numpy>=1.26.4",
+  "pillow>=11.1.0",
```
mknitt, line 26: Move this into the `dev` dependency group

mknitt: So, into the list with matplotlib and opencv in line 68

mknitt (approved):

PR #223, opened 2025-02-18
Time Slice Series Transform

### LINKED ISSUE #203: Slicing time series data
For time-dependent detection of events or faults, we need to be able to extract the timestamps from a previously recorded time series. 
The relevant features of the time series should then be cut into slices of user-specified interval length around the timestamps.

SM20598:
This is branch implements the Time Slice Series Transform. 

Closes #203


src/flowcean/polars/transforms/slice_time_series.py
```diff
@@ -0,0 +1,118 @@
...
+from flowcean.polars.is_time_series import is_timeseries_feature
+
+
+class SliceTimeSeries(Transform):
+    """Slices time series features based on a counter column.
+
+    Extracts time series segments corresponding to each timestamp in the
+    counter column. This transformation enables slicing time series data
+    into fixed-duration windows relative to reference timestamps.
+
```
ChristianWieck, line 14: I'm not quite sure what this means. Can you add a "math expression" that shows what boolean relationship needs to be true for an entry to appear in the resulting feature?

src/flowcean/polars/transforms/slice_time_series.py
```diff
@@ -0,0 +1,118 @@
...
+    | feature_a                   | feature_b                   | const |
+    | --------------------------- | --------------------------- | ----- |
+    | list[struct[time,struct[]]] | list[struct[time,struct[]]] | int   |
+    | [{12:26:01.0, {1.2}},       | [{12:26:05.0, {1.0}},       | 1     |
+    |  {12:26:02.0, {2.4}},       |  {12:26:10.0, {2.0}}]       |       |
+    |  {12:26:03.0, {3.6}},       |                             |       |
+    |  {12:26:04.0, {4.8}}]       |                             |       |
+
+    The `SliceTimeSeries` transform extracts a time window of 2 seconds
+    before each `feature_b` timestamp from `feature_a`:
```
ChristianWieck, line 27: This feels like a *very* specific use case. Can you give me an example where it might be useful for a user? I'm not sure we want this technical debt in the core flowcean framework. If it's important for your use case, maybe it's better placed in the examples folder?

src/flowcean/polars/transforms/slice_time_series.py
```diff
@@ -0,0 +1,118 @@
...
+    """
+
+    def __init__(self, counter_col: str, duration: int = 1) -> None:
+        """Initialize SliceTimeseriesTransform.
+
+        Args:
+            counter_col: The name of the counter column used to slice the time
+                series.
+            duration: The duration in seconds for slicing the time series.
+                Default is 1 second.
```
ChristianWieck, line 61: Why do you pick a default of one second? I've got the feeling this should be a "non-optional"/"required" parameter.

src/flowcean/polars/transforms/slice_time_series.py
```diff
@@ -0,0 +1,118 @@
...
+        """
+        super().__init__()
+
+        self.dur = duration
+        self.counter_col = counter_col
+
+    @override
+    def apply(self, data: pl.LazyFrame) -> pl.LazyFrame:
+        df = data.collect()
```
ChristianWieck, line 70: Do we really need to collect the frame? This is potentially *very* expensive or straight up impossible if the frame is larger-than-memory. Have a look at `LazyFrames` [`map_batches`](https://docs.pola.rs/api/python/stable/reference/lazyframe/api/polars.LazyFrame.map_batches.html).

src/flowcean/polars/transforms/slice_time_series.py
```diff
@@ -0,0 +1,118 @@
...
+        n_rows = df.height
+        remaining_time_series_columns = [
+            col
+            for col in data.columns
+            if is_timeseries_feature(data, col) and self.counter_col != col
+        ]
```
ChristianWieck, line 76: Nitpicking but do I need to / want to apply this transform to all time series features? Would it be an idea to allow the user to select them and just default to "all"?

src/flowcean/polars/transforms/slice_time_series.py
```diff
@@ -0,0 +1,118 @@
...
+        )
+
+        df = df.select(
+            pl.when(
+                pl.col(remaining_time_series_columns)
+                .struct.field("time")
+                .is_between(
+                    pl.col(self.counter_col).struct.field("time")
+                    - pl.duration(seconds=self.dur),
+                    pl.col(self.counter_col).struct.field("time"),
```
ChristianWieck, line 93: Does this also works with `float` time values? I believe we haven't fixed the time value to be `pl.DataTime` but left it up to the user.

src/flowcean/polars/transforms/slice_time_series.py
SM20598: I agree. This description is a little to specific for the transform.

src/flowcean/polars/transforms/slice_time_series.py
SM20598: I included this selection because the transform should only be applied to time series columns. Having an option to select columns for applying the transform is a nice addition, but I don't see why you would want to leave out features from this filter. Could you give me an example? :smile:

src/flowcean/polars/transforms/slice_time_series.py
SM20598: Yeah I agree. The explanation is a little shallow :+1:

src/flowcean/polars/transforms/slice_time_series.py
```diff
@@ -0,0 +1,141 @@
...
+    The SliceTimeSeries transform is called with
+    '''
+        environment.load()
+        data = environment.get_data()
+        transform = SliceTimeSeries(
+            counter_column="counter_column",
+            duration=2,
+        )
+        transformed_data = transform.transform(data)
+    '''
```
ChristianWieck, line 41:
```suggestion
    ```python
        environment.load()
        data = environment.get_data()
        transform = SliceTimeSeries(
            counter_column="counter_column",
            duration=2,
        )
        transformed_data = transform.transform(data)
    ```
```

src/flowcean/polars/transforms/slice_time_series.py
```diff
@@ -0,0 +1,141 @@
...
+    | feature_a                   | counter_column              | const |
+    | --------------------------- | --------------------------- | ----- |
+    | list[struct[time,struct[]]] | list[struct[time,struct[]]] | int   |
+    | [{12:26:03.0, {1.1}},       | [{12:26:05.0, {3.0}}],      | 1     |
+    |  {12:26:04.0, {2.2}}]       |                             |       |
+    | [{12:26:08.0, {4.4}},       | [{12:26:10.0, {6.0}}],      | 1     |
+    |  {12:26:09.0, {5.5}}]       |                             |       |
+    """
+
+    def __init__(self, counter_column: str, duration: int) -> None:
```
ChristianWieck, line 54: Why is the duration an int and not a float?

src/flowcean/polars/transforms/slice_time_series.py
```diff
@@ -0,0 +1,141 @@
...
+    def apply(self, data: pl.LazyFrame) -> pl.LazyFrame:
+        names = data.collect_schema().names()
+        counter_column_dtype = data.collect_schema()[self.counter_column]
+
+        # Collect names of all other time series columns
+        remaining_time_series_columns = [
+            col
+            for col in data.collect_schema().names()
+            if is_timeseries_feature(data, col) and self.counter_column != col
+        ]
```
ChristianWieck, line 77: Collect the schema - if possible - only once. Collecting the schema is an relative expansive operation.

src/flowcean/polars/transforms/slice_time_series.py
```diff
@@ -0,0 +1,141 @@
...
+class SliceTimeSeries(Transform):
+    """Slices time series features based on a counter column.
+
+    Extract sub-time series from features that are within a time window of n
+    seconds before each timestamp in a counter column.
+
+    feature = feature[
+        (counter_column.time - n) < time < counter_column.time
+    ]
+
```
ChristianWieck, line 17: I still don't quite get why this is needed. It feels super specific...

src/flowcean/polars/transforms/slice_time_series.py
```diff
@@ -0,0 +1,141 @@
...
+        const_columns = (
+            data.select(
+                pl.exclude(
+                    *remaining_time_series_columns,
+                    self.counter_column,
+                ),
+            )
+            .collect_schema()
+            .names()
+        )
```
ChristianWieck, line 89: Instead of collecting the schema again here, couldn't you just use the schema information you just got a few lines earlier to get the const columns names?

src/flowcean/polars/transforms/slice_time_series.py
```diff
@@ -0,0 +1,141 @@
...
+
+        # Recreate time series based on the counter column
+        data = (
+            data.unique(maintain_order=True)
+            .group_by(
+                [self.counter_column, *const_columns],
+                maintain_order=True,
+            )
+            .agg(remaining_time_series_columns)
+        )
```
ChristianWieck, line 129: Does this guarantee uniqueness? What happens if two samples have the same counter column values?

src/flowcean/polars/transforms/slice_time_series.py
ChristianWieck: I still don't quite understand the use of this transform. But since most/all other transforms offer the option to specify the target features (at least optionally), we should keep this feature here as well.

mknitt: In the robotics example, we have a long time series that includes resets of our localization state once it is delocalized. Whenever a delocalization occurs, a delocalization counter counts up. If we do not chop the time series at these resets, models would learn wrong temporal behavior around the times of these resets. That's why we need this transform. If you think this type of transform is not needed for other use cases, we can of course move it into the custom transforms for the robotics example.

ChristianWieck: Ah okay, now I get the use case. I think the current implementation is very specific - would it be an idea to have a generic "Split" transform that takes a time series and splits it into multiple pieces according to a "Timestamp" column? And a separate "LastNSeconds" (all names up for discussion 😄) transform that takes the last N seconds from a time series function? If I'm not mistaken, that would describe this transform exactly?
If you need this transform urgently, I would suggest we move it to the example as currently I lack a clear view if/how this would be usefull for another user of flowcean...

newLabAspect:
Included in #269

PR #228, opened 2025-02-20
Median transform for time series data

ChristianWieck:
This PR:
- Adds a new `Median` transform to find the mean value of a time series feature. A new feature with the name `<name>_median` is added to the dataframe.

### DIFF
--- a/src/flowcean/polars/transforms/median.py
+++ b/src/flowcean/polars/transforms/median.py
@@ -1,0 +1,34 @@
+import logging
+from collections.abc import Iterable
+
+import polars as pl
+from typing_extensions import override
+
+from flowcean.core import Transform
+
+logger = logging.getLogger(__name__)
+
+
+class Median(Transform):
+    """Replaces time-series features with there median value."""
+
+    def __init__(self, features: str | Iterable[str]) -> None:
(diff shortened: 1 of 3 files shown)


src/flowcean/polars/transforms/median.py
```diff
@@ -0,0 +1,34 @@
...
+import polars as pl
+from typing_extensions import override
+
+from flowcean.core import Transform
+
+logger = logging.getLogger(__name__)
+
+
+class Median(Transform):
+    """Replaces time-series features with there median value."""
```
schmidma, line 13:
```suggestion
    """Replaces time-series features with their median value."""
```

schmidma: it also not really replaces them, it adds them, does it?

schmidma (changes requested):

src/flowcean/polars/transforms/median.py
ChristianWieck: That's true. In the first iteration it replaced the values. I since added a `replace` argument to control the behaviour. I'll add them to the Mean and Median transforms as well.

PR #230, opened 2025-02-21
Mode transform for time series data

ChristianWieck:
This PR:
- Adds a `Mode` transform to extract the most frequent element from a time series.
- Adds a new helper script to get the value type of a time series feature.

### DIFF
--- a/src/flowcean/polars/transforms/mode.py
+++ b/src/flowcean/polars/transforms/mode.py
@@ -53,7 +53,7 @@ class Mode(Transform):
                 # Unfortunately, `mode` is not implemented for lists, so we
                 # have to use `map_elements` as a workaround.
                 .map_elements(
-                    lambda x: x.mode().item(),
+                    lambda x: x.mode().item(0, 0),
                     return_dtype=time_series_type,
                 )
             )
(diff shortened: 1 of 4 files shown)


src/flowcean/polars/transforms/mode.py
```diff
@@ -0,0 +1,63 @@
...
+                    feature,
+                )
+
+            expr = (
+                pl.col(feature)
+                .list.eval(pl.element().struct.field("value"))
+                # Unfortunately, `mode` is not implemented for lists, so we
+                # have to use `map_elements` as a workaround.
+                .map_elements(
+                    lambda x: x.mode().item(),
```
mknitt, line 56: This works only when there is one mode. For example, if the numbers 42 and 43 appear both equally often, you cannot call `.item()` since `x.mode()` would return the list `[42, 43]`.

mknitt: `polars.exceptions.ComputeError: ValueError: can only call '.item()' if the Series is of length 1, or an explicit index is provided (Series is of length 3)`

ChristianWieck: True, missed that. Now it should work. If multiple values appear the same number of times, the first one is choosen now. I know that's not "optimal", but for practical use it should be sufficent.

mknitt (approved):

PR #232, opened 2025-02-21
Remove unused workspace members

schmidma:
We currently do not use workspaces, this is a leftover from a previous attempt of using workspaces for flowcean project structure.

This PR simply removes this...


schmidma:
Superseded by #233

PR #233, opened 2025-02-21
Restructure Examples into Packages

schmidma:
Restructure examples into packages.
Having each example in its own package they can easily track their own dependencies.
Using uv workspaces (https://docs.astral.sh/uv/concepts/projects/workspaces/) allows all dependecies to be resolved at once.
Additionally this removes the need for manual tracking of dependency groups for examples, as the entire workspace can be synced using `uv sync --all-packages`


mknitt:
The examples `failure_time_predicition` and `one_tank` run on my machine after running `uv sync --all-packages`. I am not sure why they fail in the CI. 

Other than that the changes look good to me.

mknitt (approved):

PR #234, opened 2025-02-24
Add CNNs for robot_localization_failure example

mknitt:
This PR 
- adds classes for three different CNN-based classifiers
- showcases the use of CNNs in the "robot_localization_failure" example
- fixes bugs in some of the transforms that are required for the run.py


examples/robot_localization_failure/custom_transforms/particle_cloud_image.py
```diff
@@ -17,7 +17,8 @@ class ParticleCloudImage(Transform):
 
     Processes 2D particle cloud data to create grayscale images highlighting
     the particle distribution. Optionally saves images to disk and returns them
-    embedded in a Polars DataFrame for further analysis.
+    embedded in a Polars DataFrame for further analysis. The respective column
```
ChristianWieck, line 20:
```suggestion
    embedded in a Polars DataFrame for further analysis. The respective feature
```

src/flowcean/polars/transforms/match_sampling_rate.py
```diff
@@ -176,27 +176,31 @@ def interpolate_feature(
...
         cs.all(),
     )
-    # Handle scalar 'value' by wrapping into a struct
+    # Handle 'value' field
     if "value" in feature_df.columns:
-        if not isinstance(feature_df.schema["value"], pl.Struct):
-            # Wrap scalar 'value' into a struct with a single field
-            feature_df = feature_df.with_columns(
-                pl.struct([pl.col("value").alias("value")]).alias("value"),
+        is_scalar = not isinstance(feature_df.schema["value"], pl.Struct)
```
schmidma, line 181: only because a value is not a struct, it is not a scalar, is it?

src/flowcean/polars/transforms/match_sampling_rate.py
```diff
@@ -176,27 +176,31 @@ def interpolate_feature(
...
+            feature_df = feature_df.rename(
+                {"value": f"{target_feature_name}_value"},
             )
-        # Now unnest the 'value' struct
-        feature_df = feature_df.unnest("value")
+            original_field_names = [f"{target_feature_name}_value"]
+        else:
+            # Get the schema of 'value' and extract field names for structs
+            value_schema = feature_df.schema["value"]
+            if not isinstance(value_schema, pl.Struct):
```
schmidma, line 191: you do the same check here again...

src/flowcean/torch/lightning_learner.py
```diff
@@ -127,3 +131,261 @@ def configure_optimizers(self) -> Any:
             "optimizer": optimizer,
             "lr_scheduler": scheduler,
         }
+
+
+class ConvolutionalNeuralNetwork(lightning.LightningModule):
```
schmidma, line 136: we should discuss, how we want to handle creating pytorch networks as a user. Currently we find yet another layer and variant to express network architecture...

src/flowcean/torch/lightning_learner.py
```diff
@@ -127,3 +131,261 @@ def configure_optimizers(self) -> Any:
...
+    def forward(self, x: Tensor) -> Tensor:
+        for conv in self.conv_layers:
+            x = torch.relu(conv(x))
+            x = torch.max_pool2d(x, 2)
+        x = torch.flatten(x, 1)
+        for fc in self.fully_connected_layers:
+            x = torch.relu(fc(x))
+        logits = self.output_layer(x)
+        probabilities = torch.sigmoid(logits)
+        return torch.round(probabilities)
```
schmidma, line 208: why do you round here?

src/flowcean/torch/lightning_learner.py
```diff
@@ -127,3 +131,261 @@ def configure_optimizers(self) -> Any:
...
+
+    def training_step(self, batch: Any) -> Tensor:
+        inputs, targets = batch
+        outputs = self(inputs)
+        targets = targets.float()
+        return torch.nn.functional.binary_cross_entropy(outputs, targets)
```
schmidma, line 214: this is hard for the user to see. Why is this a BCE loss, and the others are MSE?

src/flowcean/torch/lightning_learner.py
```diff
@@ -127,3 +131,261 @@ def configure_optimizers(self) -> Any:
...
+        if fully_connected_layer_sizes is None:
+            fully_connected_layer_sizes = [128]
+        if dropout_rates is None:
+            dropout_rates = [0.25, 0.5]
+        self.input_shape = input_shape
+        self.save_hyperparameters()
+        self.learning_rate = learning_rate
+
+        # Convolutional layers
+        self.conv_layers = torch.nn.ModuleList()
```
schmidma, line 170: why do you use `ModuleList` here?

src/flowcean/polars/transforms/match_sampling_rate.py
mknitt: true, I changed the if statement

src/flowcean/polars/transforms/match_sampling_rate.py
mknitt: fixed

src/flowcean/torch/lightning_learner.py
mknitt: We discussed in the software meeting today and the colleagues agreed that it makes sense to keep the LightningLearner class but move the classes for the network architectures to the examples. I will do this now.

src/flowcean/torch/lightning_learner.py
mknitt: I was trying to fix an error that occurred during evaluation, since it is a binary classification task. I think we might need to apply rounding only during inference, perhaps in a separate method or post-processing step.

schmidma (approved):

PR #239, opened 2025-02-25
Cache transformed environments

ChristianWieck:
This PR:
- Adds a new `Cache` environment that allows to cache the results from other `environments` after being transformed. This should allow for a much smoother user experience as transforms only need to be applied once.
- Add a new `hash` method to `Observables`. This method returns `bytes` representing the identity of the underlying data. Environments can opt out of caching by throwing a `HashingNotSupportedError' when asked for their hash.

Limitiations:
- Currently this implementation only works for `pl.Dataframe`. However, extending it to other data types seems possible if the need arises.
- Hashing `Dataframes` is hard. The used implementation tries it best but still requires the frame to be `collected`. A different option would be to not hash the `Dataframe` but rather the source file with the `Dataframe` hash being a backup.


ChristianWieck:
Superseeded by #252 and #249.

PR #244, opened 2025-02-28
Add unnest transform

newLabAspect:
Adds an unnest transform based on the polars' `unnest` function. The transform can, e.g., be used to unnest time series data.

### DIFF
--- a/src/flowcean/polars/transforms/unnest.py
+++ b/src/flowcean/polars/transforms/unnest.py
@@ -11,7 +11,32 @@ logger = logging.getLogger(__name__)
 
 
 class Unnest(Transform):
-    """Decompose struct columns into separate columns for each field."""
+    """Decompose struct columns into separate columns for each field.
+
+    Example:
+    ```python
+    data_frame = pl.Series(
+        "c",
+        [
+            {"a": 1, "t": 1},
+            {"a": 4, "t": 2},
+            {"a": 7, "t": 3},
+            {"a": 10, "t": 4},
+            {"a": 15, "t": 5},
+        ],
+    ).to_frame()
+    ```
+    The transformed_data will be:
+    ```python
+    pl.DataFrame(
+        {
+            "a": [1, 4, 7, 10, 15],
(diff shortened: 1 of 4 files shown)


tests/polars/transforms/test_unnest.py
```diff
@@ -0,0 +1,37 @@
+import unittest
+
+import polars as pl
+from polars.testing import assert_frame_equal
+
+from flowcean.polars import Unnest
+
+
+class SelectTransform(unittest.TestCase):
```
mknitt, line 9: This should be renamed.

src/flowcean/polars/transforms/unnest.py
```diff
@@ -0,0 +1,32 @@
...
+from polars._typing import ColumnNameOrSelector
+from typing_extensions import override
+
+from flowcean.core import Transform
+
+logger = logging.getLogger(__name__)
+
+
+class Unnest(Transform):
+    """Decompose struct columns into separate columns for each field."""
```
ChristianWieck, line 14: This could use an example of how to use it and what it does. Honestly, I don't even know exactly how the dataframe changes 😄

tests/polars/transforms/test_unnest.py
mknitt, line 1: I think we need a test for our time series format with multiple columns. I think the current implementation of the `Unnest` class does not support when for example `time` occurs more than once. Or maybe this is not the scope of this transform, and we need an `UnnestTimeSeries` transform?

tests/polars/transforms/test_unnest.py
newLabAspect: Done

tests/polars/transforms/test_unnest.py
newLabAspect: That would be a nice thing to have. Anyways, I would put that rather in a separate PR taking this PR just to add a wrapper for the polars unnest.

ChristianWieck:
Lgtm, @mknitt I leave approval up to you, as you have open discussions left.

mknitt (approved):

PR #246, opened 2025-03-05
Add Sliding Window for Timeseries

ChristianWieck:
This PR:
- Adds a `TimeSeriesSlidingWindow` transform. This transform is not very efficient, but I think we would rather have an inefficient transform than none at all. But as I'm using `map_batches` this should be compatible with the streaming engine and does not require all data to be present in memory at the same time.

The transform could still use an example. I'll add one, once we agree on the generel functionality.

### DIFF
--- a/src/flowcean/polars/is_time_series.py
+++ b/src/flowcean/polars/is_time_series.py
@@ -3,22 +3,30 @@ from typing import cast
 import polars as pl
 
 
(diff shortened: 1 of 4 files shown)


mknitt (approved):
Looking at the tests, I initially thought that it might only work for scalar time series, but I tested it for lists and dictionaries and it works fine. The code looks good to me as well.

PR #247, opened 2025-03-07
Update locked dependencies

ChristianWieck:
This PR:
- Update the `uv.lock` dependencies to their latest versions. This is especially useful for `polars`, as the new versions offer better performance, and `torch`, as the current locked version has problems on Windows machines, but it doesn't hurt any of the other dependencies either.


schmidma (approved):
ok, if the CI is happy, i'm happy ;)

PR #249, opened 2025-03-13
Add generic Transform hashing

ChristianWieck:
This PR:
- Adds a mechanism to compute a hash from arbitrary transforms.
- Adds `cloudpickle` as a dependency for pickle lambdas.
- Adds two tests for hashing. If this PR get's merged we might want to extend the test suite.

This PR is intended to be a "catch-all" with no extra work for transform developers. However, this leads to extreme sensitivity about the order of arguments, even if the order doesn't matter. It should keep this:

- If two hashes are the same, the transforms behave the same
- If two hashes are *not* the same, the transforms _might_ behave differently - but we cannot be sure.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -18,7 +18,7 @@ keywords = []
 authors = [{ name = "The AGenC Team", email = "agenc.team@tuhh.de" }]
 
 dependencies = [
-  "dill>=0.3.9",
+  "cloudpickle>=3.1.1",
   "docker>=7.1.0",
   "grpcio>=1.67.0",
   "joblib>=1.4.2",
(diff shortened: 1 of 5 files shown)


mknitt (approved):
The "type:ignore" comments look wrong at first, but I think cloudpickle just does not support type hinting, yet. 
The rest of the code looks good to me.

ChristianWieck:
> The "type:ignore" comments look wrong at first, but I think cloudpickle just does not support type hinting, yet. The rest of the code looks good to me.

That's exactly the case. We could potentially add our own type stubs, but tbh I'm not sure we should.

ChristianWieck:
The merge queue fails for this PR. @schmidma do you have an idea why this happens?

PR #251, opened 2025-03-13
Constraint polars to a version before 1.23

### LINKED ISSUE #250: Pin polars version < 1.23
When running flowcean test code polars throws the following warning:

> The old streaming engine is being deprecated and will soon be replaced by the new streaming engine. Starting Polars version 1.23.0 and until the new streaming engine is released, the old streaming engine may become less usable. For people who rely on the old streaming engine, it is suggested to pin your version to before 1.23.0.
More information on the new streaming engine: https://github.com/pola-rs/polars/issues/20947

We are currently using the old streaming engine and should therefore pin a version prior to 1.23.0.

ChristianWieck:
This PR:
- See title and #250.

Closes #250 .

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -23,7 +23,7 @@ dependencies = [
   "joblib>=1.4.2",
   "lightning>=2.4.0",
   "numpy>=1.26.4",
-  "polars>=1.9.0",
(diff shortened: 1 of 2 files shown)


ChristianWieck:
Superseded by #255 .

PR #264, opened 2025-04-09
Add Prediction Tool

### LINKED ISSUE #257: Define "cps-adapter" interface and process loop
To connect CPS to flowcean we need to define:
- The adapter interface which is required to communicated with the actual CPS via some sort of protocol (ROS, OPC-UA, ...)
- Define the process loop. This is closely related to the tools, but not the same. Essentially it boils down to whether we want a pull or push approach to requests.

ChristianWieck:
This PR:
- Adds a new `Adapter` class providing an interface to connect CPS with Flowcean in Real-Time.
- Adds a prediction tool loop, to run a `Model` with an `Adapter` in a closed loop.

Closes #257.

### DIFF
--- a/src/flowcean/core/adapter/__init__.py
+++ b/src/flowcean/core/adapter/__init__.py
@@ -1,0 +1,6 @@
+__all__ = [
+    "Adapter",
+    "Stop",
+]
+
+from .adapter import Adapter, Stop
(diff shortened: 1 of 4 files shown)


newLabAspect (approved):
lgtm

PR #265, opened 2025-04-09
Add xor example

ChristianWieck:
This PR:
- Adds a simple `xor` example. This example demonstrates how a learner can be trained and used with the prediction loop and the `DataFrameAdapter`.

Based on #264


src/flowcean/polars/adapter/dataframe_adapter.py
```diff
@@ -0,0 +1,54 @@
...
+from typing import cast
+
+import polars as pl
+
+from flowcean.core.adapter import Adapter
+from flowcean.core.adapter.adapter import Stop
+from flowcean.polars.environments.dataframe import DataFrame
+
+
+class DataFrameAdapter(Adapter):
```
newLabAspect, line 11: Do we need tests for this class?

newLabAspect (commented):
Should we add this example in the justfile and the examples in the pipeline?

ChristianWieck:
> Should we add this example in the justfile and the examples in the pipeline?

I'm not sure how the pipeline likes to store a model and its outputs. And most/all functionality should already be covered by the unit tests, so we would not really gain much. But I'm open to it.

src/flowcean/polars/adapter/dataframe_adapter.py
ChristianWieck: Would be better. I'll have a look how easily a test can be written.

ChristianWieck: I added a simple test for the `DataFrameAdapter`.

newLabAspect:
An advantage of having the example in the pipeline would be to make sure that we do not brake the example through changes in the functionalities. This is probably a good thing if we assume future changes in the adapter and tool structures.

ChristianWieck:
> An advantage of having the example in the pipeline would be to make sure that we do not brake the example through changes in the functionalities. This is probably a good thing if we assume future changes in the adapter and tool structures.

Alright. I added the example to the just file and the ci/cd tasks.

newLabAspect (approved):

PR #266, opened 2025-04-09
Allow path string in model save or load

### LINKED ISSUE #263: Support string paths in Model load and save
The `Load` and `Save` method from `Model` should support string paths as well.

ChristianWieck:
This PR:
- Changes the `Model` `Save` and `Load` method to accept a string path.

Closes #263.


newLabAspect (approved):
lgtm

PR #281, opened 2025-05-28
Fix Example Dependencies and Python Versions

ChristianWieck:
This PR:
- Makes the `flowcean` dependency for all examples explicit.
- Reduces the allowed python version for the `energy-system` example down to `3.12`.
- Upgrades the lock file (required by checks) and makes #279 obsolete.


stbalduin (approved):
Changes are all straight forward. This can be merged.

PR #300, opened 2025-06-13
Add Discrete Derivative Transform

ChristianWieck:
This PR:
- Adds a `DiscreteDerivative` Transform with tests. The transform supports `forward`, `backward` and `central` derivatives.


src/flowcean/polars/transforms/discrete_derivative.py
```diff
@@ -0,0 +1,116 @@
...
+from typing import Literal, TypeAlias
+
+import polars as pl
+from typing_extensions import override
+
+from flowcean.core import Transform
+
+logger = logging.getLogger(__name__)
+
+DiscreteDerivativeType: TypeAlias = Literal["forward", "backward", "central"]
```
schmidma, line 12:
```suggestion
DiscreteDerivativeKind: TypeAlias = Literal["forward", "backward", "central"]
```
I don't know what the convention is in Python, but I feel like this is not a _type_

src/flowcean/polars/transforms/discrete_derivative.py
```diff
@@ -0,0 +1,116 @@
...
+        self.method = method
+
+    @override
+    def apply(self, data: pl.LazyFrame) -> pl.LazyFrame:
+        feature_names = data.collect_schema().names()
+        for feature in self.features:
+            value_feature = f"{feature}_value"
+            time_feature = f"{feature}_t"
+            dt_feature = f"{feature}_dt"
+            dvalue_feature = f"{feature}_derivative"
```
schmidma, line 47: make this an configurable argument

src/flowcean/polars/transforms/discrete_derivative.py
```diff
@@ -0,0 +1,116 @@
...
+                        ).alias(dt_feature),
+                    ],
+                )
+            else:
+                logger.warning("Unknown derivative method %s", self.method)
+
+            # Calculate the derivative
+            data = data.filter(
+                pl.col(value_feature).is_not_null(),
+            ).with_columns(
```
schmidma, line 94: I think there is a drop null or similar?

src/flowcean/polars/transforms/discrete_derivative.py
```diff
@@ -0,0 +1,116 @@
...
+                        pl.col(feature)
+                        .list.eval(pl.first().struct.field("value"))
+                        .alias(value_feature),
+                        pl.col(feature)
+                        .list.eval(pl.first().struct.field("time"))
+                        .alias(time_feature),
+                    ],
+                )
+                .with_row_index(name=index_feature)
+                .explode([value_feature, time_feature])
```
schmidma, line 62: this still contains all the features in the data frame, does it? Can we select beforehand? only using the features for the derivative?

src/flowcean/polars/transforms/discrete_derivative.py
```diff
@@ -0,0 +1,116 @@
...
+            time_feature = f"{feature}_t"
+            dt_feature = f"{feature}_dt"
+            dvalue_feature = f"{feature}_derivative"
+            index_feature = "_index"
+
+            data = (
+                data.with_columns(
+                    [
+                        pl.col(feature)
+                        .list.eval(pl.first().struct.field("value"))
```
schmidma, line 54: huh? this part i don't get... why taking the first?

src/flowcean/polars/transforms/discrete_derivative.py
ChristianWieck: There is, thanks. I changed it.

PR #301, opened 2025-06-16
Alp Research

COOOkies4U:
The objective of this research is the Automatic Lashing Platform (ALP), a novel technical solution developed by _KALP GmbH_ for the automatic insertion and removal of twistlocks during waterside container handling. The ALP operates using a self-sufficient hydraulic system that requires no external energy supply. This research addresses the key challenge of estimation the container weight, which is critical for optimizing the valve adjustments of the hydraulic system  and maximizing energy recovery. This project investigates the data-driven generation of models for Cyber-Physical Systems (CPS), focusing on the integration of a dedicated framework called Flowcean into the development workflow of the ALP. This case study aims to improve and evaluate the capabilities of Flowcean to generate a predictive model for the ALP. The findings demonstrate that Flowcean is capable of generating a model to estimate the container weight by analyzing the pressure curve of the system using a decision tree model.


COOOkies4U:
I removed the uvloop package from the uv.lock file for the following reason:

_RuntimeError: uvloop does not support Windows at the moment
help: `uvloop` (v0.21.0) was included because `automatic-lashing-platform` (v0.1.0) depends on `flowcean` (v0.4.0) which depends on `palaestrai` (v3.5.8) which depends on `uvloop`_

src/flowcean/polars/transforms/derivative.py
ChristianWieck, line 1: I've got the feeling that this transform is pretty specific. I tried to do something similar and more general with #300. Could we use those transforms? If we keep this transform in the upstream flowcean it would require some docstrings to explain what it really does and how to use it...

examples/automatic_lashing_platform/run.py
ChristianWieck, line 1: Their should be a flag to save the model as a `.fml`. This makes importing and using it after training significantly easier.

ChristianWieck (approved):

PR #317, opened 2025-07-24
Use `collect_all` instead of two separate collects

ChristianWieck:
This MR:
- Replaces some `collect` calls with `pl.collect_all`. The later allows polars to apply optimizations to the whole query and reuse cached results. Those changes could potentially speed up learning by a good amount.


newLabAspect (approved):
lgtm

PR #330, opened 2025-09-08
Update UV  and Just to Latest Versions in CI/CD

ChristianWieck:
This PR:
- See title.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,7 +21,6 @@ jobs:
       - name: Set up uv
         uses: astral-sh/setup-uv@v6
         with:
-          version: "latest"
           prune-cache: false
           cache-local-path: "${{ runner.workspace }}/.cache/uv"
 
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -19,14 +19,14 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Set up uv
-        uses: astral-sh/setup-uv@v3
+        uses: astral-sh/setup-uv@v6
         with:
-          version: "0.5.1"
+          version: "latest"
```
schmidma, line 24: We should just drop that, as specified in the documentation:
> If you do not specify a version, this action will look for a [required-version](https://docs.astral.sh/uv/reference/settings/#required-version) in a uv.toml or pyproject.toml file in the repository root. If none is found, the latest version will be installed.

schmidma (changes requested):

.github/workflows/ci.yml
ChristianWieck: Fair, I'll remove it.

schmidma (approved):

PR #344, opened 2025-09-08
Detect input/output size in MultiLayerPerceptron

schmidma:


src/flowcean/torch/lightning_learner.py
```diff
@@ -53,6 +52,14 @@ def learn(
...
     ) -> PyTorchModel:
+        # Infer input_size and output_size from the data
+        input_size = len(inputs.collect_schema().names())
+        output_size = len(outputs.collect_schema().names())
+
+        # If the module is a MultilayerPerceptron, initialize its layers
+        if isinstance(self.module, MultilayerPerceptron):
+            self.module.initialize_layers(input_size, output_size)
+
         dfs = pl.collect_all([inputs, outputs])
```
ChristianWieck, line 63: We collect the input and output frame here anyways - we could use those schemas to get the input and output size.

PR #355, opened 2025-09-09
Add select best model method to report

mknitt:


src/flowcean/core/report.py
```diff
@@ -135,6 +137,68 @@ def pretty_print(
             )
             console.print(panel)
 
+    def select_best_model(
+        self,
+        metric_name: str,
+        *,
+        mode: str = "maximize",
```
schmidma: This should use the `Literal` type-hint. To show what options for `mode` are allowed

src/flowcean/core/report.py
```diff
@@ -135,6 +137,68 @@ def pretty_print(
...
+    ) -> Model:
+        """Return the best model according to a metric.
+
+        Args:
+            metric_name: Name of the metric to select by.
+            mode: One of {"maximize", "minimize", "max", "min"}.
```
schmidma: Decide on long or short version, do not support both

src/flowcean/core/report.py
```diff
@@ -135,6 +137,68 @@ def pretty_print(
...
+
+        Returns:
+            The best Model instance. Requires that a mapping of models was
+            attached to the report under `models_by_name` (as done by
+            evaluate_offline).
```
schmidma: This is exactly the inverse requirement. The report should either carry this information always, or the function requires this mapping as an argument

src/flowcean/core/report.py
```diff
@@ -135,6 +137,68 @@ def pretty_print(
...
+        if mode_norm in ("maximize", "max"):
+            factor = 1.0
+        elif mode_norm in ("minimize", "min"):
+            factor = -1.0
+        else:
+            msg = "mode must be one of: maximize, minimize, max, min"
+            raise ValueError(msg)
+
+        best_name: str | None = None
+        best_score = float("-inf")
```
schmidma: This is https://docs.python.org/3/library/functions.html#max

src/flowcean/core/report.py
```diff
@@ -135,6 +137,68 @@ def pretty_print(
...
+        for model_name, entry in self.items():
+            if metric_name not in entry:
+                continue
+            value = entry[metric_name]
+            if isinstance(value, Mapping):
+                value = cast("Mapping[str, Reportable]", value)
+                vals = [float(str(v)) for v in value.values()]
+                if not vals:
+                    continue
+                score = sum(vals) / len(vals)
```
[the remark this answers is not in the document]
mknitt: Added a note in the docstring explaining that nested metrics use arithmetic mean.

src/flowcean/core/strategies/offline.py
```diff
@@ -103,4 +105,78 @@ def evaluate_offline(
...
+    raw = entry[metric_name]
+    if isinstance(raw, Mapping):
+        # Average over sub-metrics
+        mapped = cast("Mapping[str, Reportable]", raw)
+        values = [_to_float(v) for v in mapped.values()]
+        return None if not values else sum(values) / len(values)
+    return _to_float(raw)
+
+
+def select_best_model(
```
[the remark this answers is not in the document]
mknitt: removed

PR #365, opened 2025-09-17
Track Input Features for Regression Trees

ChristianWieck:
This PR:
- Adds additional tracking of the input feature names for the regression tree. Without this change the regression tree rejects observation with too many features, even if all required features are present.

### DIFF
--- a/src/flowcean/sklearn/model.py
+++ b/src/flowcean/sklearn/model.py
@@ -23,6 +23,7 @@ class SciKitModel(Model):
     """A model that wraps a scikit-learn model."""
 
     estimator: SupportsPredict
+    input_names: list[str] | None
     output_names: list[str]
 
     def __init__(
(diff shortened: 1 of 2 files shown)


src/flowcean/sklearn/model.py
```diff
@@ -53,6 +60,9 @@ def _predict(
         if isinstance(input_features, pl.LazyFrame):
             input_features = input_features.collect()
 
+        if self.input_names is not None:
+            input_features = input_features.select(self.input_names)
```
schmidma, line 64: better first select, then collect

schmidma (changes requested):
I'm not sure about this change. This may be something general, we have to quickly discuss. Do we want to support that for all our models? Having them accept a superset of features, picking the ones it needs? Throwing an error could also be considered a feature.

src/flowcean/sklearn/model.py
ChristianWieck: Good point. We should probably discuss this in the team. Either all models should select their respective inputs and discard other features - or raise an error if they encounter features they didn't train on.

ChristianWieck:
We plan for models to have knowledge about their inputs and outputs. In this context, the input features should also be checked. This PR will therefore be replaced by a broader one that will introduce a check for all models.

PR #402, opened 2026-06-04
Example windturbine

uenge:
First version of a wind turbine example as described by Schuler et al.
Some annotation of parameters are still missing in the comments.
Discrete modes currently divided by if statements (instead of locations).


schmidma:
Thanks @uenge for the original wind-turbine proposal! Closing this in favor of the follow-up in #432.

